PR #10, opened 2020-12-22
Use `frozen_string_literals`

dixpac:
To match other Rails engines


dixpac:
@dhh I've fixed the conflicts. When you have some time just let me know do you want this in. If not, I will close so we don't "bloat" open PRs list 😃

PR #23, opened 2020-12-23
doc: Add Torbo.start() to the installation guide

imWildCat:
After following the installation guide, I found the counterpart of `Turbolinks.start()` is `Turbo.start()`.

It's better to have it in the guide for clarity.


javan:
Turbo [starts itself now](https://github.com/hotwired/turbo/blob/35ce4aacdd4397092e9bbba682272ae148c8c4d0/src/index.ts#L6-L7)! :)

Issue #19, opened 2020-12-23
Create Webpacker specific installer


dhh:
The `turbo:install` command was designed for the asset pipeline. Let's create two new commands: `turbo:install:asset_pipeline` and `turbo:install:webpacker`. Then we'll make `turbo:install` delegate to the right one, depending on whether Webpacker is active in the app.

The `turbo:install:webpacker` command should make the following changes:

1. Remove turbolinks and @rails/ujs from the package.json
2. Remove those libraries from the default application.js pack
3. Include @hotwired/turbo-rails as described in the README in the pack file.

dhh:
Fixed by #27

PR #27, opened 2020-12-24
Create Webpacker specific install.

dixpac:
Introduce 3 new rake tasks:

* `turbo:install` delegates to webpacker if it is active, else uses asset
pipeline.
* `turbo:install:webpacker` hooks up webpacker
* `turbo:install:asset_pipeline` hooks up asset pipeline

references #19

### DIFF
--- a/lib/tasks/turbo_tasks.rake
+++ b/lib/tasks/turbo_tasks.rake
@@ -1,3 +1,5 @@
+def run_install_template(path) system "#{RbConfig.ruby} ./bin/rails app:template LOCATION=#{File.expand_path("../install/#{path}.rb",  __dir__)}" end
+
 namespace :turbo do
   desc "Install Turbo into the app"
   task :install do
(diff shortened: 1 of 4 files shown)


lib/tasks/turbo_tasks.rake
```diff
@@ -1,6 +1,22 @@
 namespace :turbo do
-  desc "Install Turbo into the app"
+  desc "Install Turbo into the app. If webpacker is configured delegates to webpacker, else uses asset pipline"
   task :install do
-    system "#{RbConfig.ruby} ./bin/rails app:template LOCATION=#{File.expand_path("../install/turbo.rb", __dir__)}"
+    if File.exist? 'bin/webpack'
```
dixpac, line 4: Maybe there is a better way to check for `webpacker` ?

swandive: Perhaps ```!!defined?(Webpacker)```

dixpac:
@dhh if you need me to fix/change something on this PR, just let me know ✌🏼

PR #42, opened 2020-12-26
🔒 Fix missing HTTPS on link

jbampton:


dhh:
This is a generated file just included as a test dummy. If this is present in Rails still, that’s where to patch 👍

PR #44, opened 2020-12-27
add example to send multiple actions down the turbo_stream

udkl:
Adding an example that isn't obvious from the `turbo_stream` API


dhh:
Know this is possible but I’d consider it an anti pattern. Any multi action stream should use a full template.

seanpdoyle:
> Any multi action stream should use a full template.

I've opened https://github.com/hotwired/turbo-site/pull/14 to emphasize Rails' custom format template support.

PR #45, opened 2020-12-27
Document `turbo_stream.erb` templates

seanpdoyle:
Related to #44

Elaborate on the ways that a Rails controller can send back a
`turbo_stream` response, and demonstrate how to render multiple streams
as part of a single template.

### DIFF
--- a/app/helpers/turbo/streams_helper.rb
+++ b/app/helpers/turbo/streams_helper.rb
@@ -1,6 +1,32 @@
 module Turbo::StreamsHelper
   # Returns a new <tt>Turbo::Streams::TagBuilder</tt> object that accepts stream actions and renders them them as
   # the template tags needed to send across the wire. This object is automatically yielded to turbo_stream.erb templates.
+  #
+  # When responding to HTTP requests, controllers can declare `turbo_stream` format response templates in that same
+  # style as `html` and `json` response formats. For example, consider a `MessagesController` that responds to both
+  # `text/html` and `text/html; turbo-stream=*; charset=utf-8` requests along with a `.turbo_stream.erb` action template:
+  #
+  #   def create
+  #     @message = Message.create!(params.require(:message).permit(:content))
+  #     respond_to do |format|
+  #       format.turbo_stream
+  #       format.html { redirect_to messages_url }
+  #     end
+  #   end
+  #
+  #   <%# app/views/messages/create.turbo_stream.erb %>
+  #   <%= turbo_stream.append "messages" do %>
+  #     <%= render @message %>
+  #   <% end %>
+  #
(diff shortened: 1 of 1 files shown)


app/helpers/turbo/streams_helper.rb
```diff
@@ -1,6 +1,28 @@
...
+  #
+  # <%# app/views/messages/create.turbo_stream.erb %>
+  # <%= turbo_stream.append(:messages, partial: "messages/message",
+  #       locals: { message: @message }) %>
+  #
+  # When a `app/views/messages/_message.turbo_stream.erb` template exists, the
+  # controller will respond to requests to the `MessagesController#create` that
+  # expects a `text/html; turbo-stream=*; charset=utf-8` response by rendering the
+  # `messages/create.turbo_stream.erb` view template and transmitting the response
+  # over the WebSocket connection.
```
udkl, line 25: `transmitting the response over the WebSocket connection`

AFAIK, there is no `WebSocket` connection. My app was receiving plain HTTP responses that contained streams.

app/helpers/turbo/streams_helper.rb
```diff
@@ -1,6 +1,28 @@
...
   def turbo_stream
     Turbo::Streams::TagBuilder.new(self)
   end
```
udkl, line 28: transmitting the response over the WebSocket connection

AFAIK, there is no WebSocket connection. My app was receiving plain HTTP responses that contained streams.

app/helpers/turbo/streams_helper.rb
```diff
@@ -1,6 +1,28 @@
...
+  #   respond_to do |format|
+  #     format.turbo_stream
+  #     format.html { redirect_to messages_url }
+  #   end
+  # end
+  #
+  # <%# app/views/messages/create.turbo_stream.erb %>
+  # <%= turbo_stream.append(:messages, partial: "messages/message",
+  #       locals: { message: @message }) %>
+  #
```
udkl, line 20: While you are at it, could you also add an example for multiple frames in the same message ? :

```erb
# It is also possible to render multiple turbo stream actions : 
# <%# app/views/messages/create.turbo_stream.erb %>
# <%= turbo_stream.append "messages" do %>
#  <%= render @message %>
# <% end %>
#
# <%= turbo_stream.replace "new_message" do %>
#  <%= render partial: "new_message", locals: {chatroom_id: @chatroom.id} %>
# <% end %>
```

app/helpers/turbo/streams_helper.rb
seanpdoyle: Good catch, thank you!

PR #51, opened 2020-12-28
Put app in autoload_once_paths

dhh:


lib/turbo/engine.rb
```diff
@@ -6,6 +6,15 @@ class Engine < Rails::Engine
     isolate_namespace Turbo
     config.eager_load_namespaces << Turbo
     config.turbo = ActiveSupport::OrderedOptions.new
+    config.autoload_once_paths = %W(
```
kaspth, line 9: Changed to `+=` here https://github.com/hotwired/turbo-rails/commit/d83a821307aa1ac5cd6616cd5626e80149439fac

PR #53, opened 2020-12-29
Mixin ActionView::RecordIdentifier#dom_id

seanpdoyle:
When the `Turbo::Streams::TurboStreamsTagBuilder` module is mixed-into a
Controller, also mix in the [ActionView::RecordIdentifier#dom_id][dom_id].

[dom_id]: https://edgeapi.rubyonrails.org/classes/ActionView/RecordIdentifier.html#method-i-dom_id

### DIFF
--- a/app/controllers/turbo/streams/turbo_streams_tag_builder.rb
+++ b/app/controllers/turbo/streams/turbo_streams_tag_builder.rb
@@ -16,6 +16,8 @@
 module Turbo::Streams::TurboStreamsTagBuilder
   private
 
+  delegate :dom_id, to: ActionView::RecordIdentifier
+
   def turbo_stream
     Turbo::Streams::TagBuilder.new(view_context)
   end


dhh:
Why?

seanpdoyle:
> Why?

@dhh the documentation [_used to_ contain example code demonstrating how to render with `turbo_stream.replace` and a `dom_id`](https://github.com/hotwired/turbo-site/pull/11#discussion_r548322992).

[It no longer does](https://github.com/hotwired/turbo-site/blob/a1e470fd672e192497b71d3ad1b4d89905fb69f7/docs/handbook/04_streams.md#reusing-server-side-templates), but it can be surprising when moving a `turbo_stream` response template between an `app/views/` file and a `respond_to` block in an `app/controllers` file.

An alternative to including the delegation declaration could be to mention this quirk in the docs somewhere, either as part of the Ruby docs or the https://github.com/hotwired/turbo-site docs.

dhh:
You're going to get the dom_id for free when you use a template, but yes, not in the controller. I don't think using explicit dom_id is used anyway. We should just take a dom_id able object and call it ourselves.

PR #55, opened 2020-12-29
Slim support for rails turbo:install to insert lines to application.html.slim

equivalent:
related to https://github.com/hotwired/hotwire-rails/issues/6 

```
Given I'm using `slim-rails` 
And I have `app/views/application.html.slim` as a layout file
When I `rails hotwire:install`
Then I generator should insert  required lines to my layout file
```

Original

Screenshot from 2020-12-29 20-51-18

Generator
Screenshot from 2020-12-29 20-47-08

After success

Screenshot from 2020-12-29 20-51-42

### Related links

* [what is Slim](http://slim-lang.com/)
* [Slim Rails](https://github.com/slim-template/slim-rails) 

## Why you should consider merging this PR

* To support other template llib not just default ERB
* ease of implementation for junior  developers using Slim 

## Why you should NOT consider merging this PR

I originally thought I'm helping  with this PR, but more I've tried it out I've realized I'm not.  

Now I agree that  the warning message ` Add <%= turbo_include_tags %> and <%= yield :head %> within the <head> tag in your custom layout.` is efficient enough as a solution.

My PR may not be the best idea as it's hard to to determine the end of the `head` tag in  [Slim](http://slim-lang.com/) therefore I'm inserting it bellow  occurrence of  the word `\s+head` which may be really bad idea if the project name is something like `title shot to the head` or something 

other solution would be to insert it bellow `= csrf_meta_tags` or something (what do you think ?)

Anyway I'm not recommending to merge this in as it is. However I would like to hear  your opinion on if you find any variation of this PR s useful so I can alter it; ...or feel free to close it. (At this point I'm 90% sure this is a useless  PR)

### DIFF
--- a/lib/install/turbo.rb
+++ b/lib/install/turbo.rb
@@ -1,12 +1,29 @@
-APPLICATION_LAYOUT_PATH = Rails.root.join("app/views/layouts/application.html.erb")
+APPLICATION_LAYOUT_PATH_ERB = Rails.root.join("app/views/layouts/application.html.erb")
+APPLICATION_LAYOUT_PATH_SLIM = Rails.root.join("app/views/layouts/application.html.slim")
 
-if APPLICATION_LAYOUT_PATH.exist?
+layout_file_exist = false
+
+if APPLICATION_LAYOUT_PATH_ERB.exist?
   say "Yield head in application layout for cache helper"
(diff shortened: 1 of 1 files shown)


dhh:
Appreciate you taking a stab at it. I think it's too much. And it's going to be more complex adding for HAML as well. I think we can maybe tweak the error/warning message to recognize other template processors.

PR #71, opened 2021-01-05
Document how to use turbo-rails with native adapters

ghiculescu:
The [Android](https://github.com/hotwired/turbo-android/blob/8fb001a4d20bbb0517e554fa4ba18c1969f5722d/turbo/src/main/assets/js/turbo_bridge.js#L10) and [iOS](https://github.com/hotwired/turbo-ios/blob/210f83a115e2deb113564b0d31bca0ed3aeddede/Source/WebView/turbo.js#L11) adapters both require `window.Turbo`.

If you just do `import "@hotwired/turbo-rails"` nothing is set on the `window` (this is a change from Turbolinks). So you need to set it manually if you want to use a native adapter. This PR adds a note explaining this to the docs.

(For the record, I think it would be better to just assign `window.Turbo` always - it would also simplify the current instructions around when you should/shouldn't import the `Turbo` instance. Happy to look into a PR for that as an alternative.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,14 @@ The JavaScript for Turbo can either be run through the asset pipeline, which is 
 
 Running `turbo:install` will install through NPM if Webpacker is installed in the application. Otherwise the asset pipeline version is used.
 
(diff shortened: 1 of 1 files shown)


dhh:
I agree we `window.Turbo`. Any further thoughts @sstephenson or @javan?

ghiculescu:
> I agree we `window.Turbo`. Any further thoughts @sstephenson or @javan?

As in, you agree turbo should always set a window prop? I can make a PR if the other guys agree.

PR #75, opened 2021-01-08
Change Turbo Stream media type

dunglas:
Follows hotwired/turbo#91.

### DIFF
--- a/app/helpers/turbo/streams_helper.rb
+++ b/app/helpers/turbo/streams_helper.rb
@@ -4,7 +4,7 @@ module Turbo::StreamsHelper
   #
   # When responding to HTTP requests, controllers can declare `turbo_stream` format response templates in that same
   # style as `html` and `json` response formats. For example, consider a `MessagesController` that responds to both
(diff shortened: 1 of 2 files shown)


dunglas:
Updated to use the media type introduced in hotwired/turbo#93 (cc @sstephenson)

Intrepidd:
👍  turbo beta 3 got released today so we need this for it to work properly on rails

dixpac:
I can confirm that this is an issue I'm seeing this morning. Altho, I'm not sure wether https://github.com/hotwired/turbo-rails/blob/main/app/assets/javascripts/turbo.js requires any changes also

Intrepidd:
Thanks @dhh 🙏 

Could we get a release though, AFAICS latest turbo-rails released version does not work properly with latest turbo beta. Thanks !

dhh:
https://rubygems.org/gems/turbo-rails/versions/0.5.4 includes this fix. If you've found a bug with that, do open an issue on how to reproduce 👍

Intrepidd:
My bad all is well on the turbo-rails side :+1:

Issue #22, opened 2020-12-23
Response fails to render (client-side) if view files don't contain `.html` in the name


cmartyn:
We have a relatively large Rails codebase that uses the slim templating language along with Turbolinks (including native Turbolinks wrappers for iOS and Android). In general, it has worked great for us for years! (Thanks!)

But when I upgraded to Turbo, all our links appeared to stop working: The URL would change, and the client would make the request to the server (per browser console), and the server would render a response (per the Rails logs), but the DOM would never be updated with the content rendered by the server.

After a lot of trial and error, I finally figured out why: Our view files are generally named `file_name.slim` rather than `file_name.html.slim`. There was one link I could find that *did* work, and it happened to be one of the few files that included the `.html` extension in its name.

Using this naming convention for our view files hasn't caused us issues before, so it's technically a regression from Turbolinks 5. I'm not sure if this is a common pattern that could affect lots of people, or if we're the only app doing it and we've been doing it wrong for years.

I figured it was worth posting here to both clarify our upgrade path (should we bite the bullet and do a massive file rename, or will this be treated as a bug to be fixed?) and to save others time if they encounter the same issue.

dhh:
That's a good catch. It's caused by the need to switch rendering formats when streams are rendered, see https://github.com/hotwired/turbo-rails/blob/6ba6e005990682a61d82067dc141afdc98bd6c22/app/channels/turbo/streams/broadcasts.rb#L26. Can only force that switch if the mime type is present in the template.

Not sure the best way to catch/alert about this. Maybe there's a way to catch the exception in https://github.com/hotwired/turbo-rails/blob/6ba6e005990682a61d82067dc141afdc98bd6c22/app/channels/turbo/streams/broadcasts.rb#L63-L65 and rethrow with better error messaging? Please do investigate!

dvisockas:
@cmartyn, after a weekend of digging found a rather hacky fix: remove 'turbo-stream;' from `Content-Type` headers: 
```ruby
if response.headers['Content-Type'].include? 'text/html'
  response.headers['Content-Type'].remove!('turbo-stream;')
end
```

Issue #52, opened 2020-12-28
Link_to with method: :post redirects


arpansac:
Just found these:

- If we use the `link_to` tag with `method: :post` it redirects instead of rendering inside the turbo_frame (get works fine)
whereas `button_to' displays usual behaviour by rendering it inside the turbo_frame

- `link_to...do` always redirects to a new page

Are these the usual behaviors, I'm totally new, watching a lot of tutorials here

dhh:
The remote-powered `link_to` helpers do not currently trigger Turbo. I'll document this, and then we'll consider whether we'll want to support this in the future. Thanks for reporting!

dhh:
Documented in 9d53529bedd4de19dea8521d415f40685dfe7749.

PR #80, opened 2021-01-11
Templates without .html aren't rendered

dvisockas:
As pointed in #22 , there is a case when a view has no `.html` extension (`.slim` instead of `.html.slim`)
While inspecting the request/response lifecycle in the browser I observed that the first load of the page doesn't have the `turbo-stream` mime type in the Accept header, however when the JS is loaded, subsequent requests have that header value present. 
Somewhere in the library code, this causes the `turbo:before-fetch-response` event called from `FetchRequest#receive` JS class to be `preventDefault()-ed`.
If the template has a `.html` extension - response ceases to include `turbo-stream` in the `Content-Type` header
Currently, the PR removes `.html` extension from `views/trays/index(.html.erb -> .erb)` and adds the Accept turbo-stream header request to the test suite. 
Would be happy to make the test green, however after going through all the ruby and JS codebase I cannot figure out where the hiccup might happen.


priit:
I had same issue with haml templates without html prefix in file name. 

I made very quick hack to study turbo-rails futher and forced content_type into text/html application_controller.rb:

`  after_action { self.content_type = "text/html; charset=utf-8" if request.env["HTTP_ACCEPT"] == "text/vnd.turbo-stream.html, text/html, application/xhtml+xml" }
`
Now I turbo_frame_tag works with GET request as expected.

dvisockas:
Seems to be fixed by [#52](https://github.com/hotwired/turbo/pull/52)

PR #82, opened 2021-01-16
Fix: Erb typo

dixpac:
Fixes, small typo on the `TagBuilder` docs.


dixpac:
I've scanned through all of the doc examples including `<% end %>` and this is the only one I found

PR #83, opened 2021-01-16
Adding ability to broadcast to multiple streams for has_many and has_and_belongs_to_many relationships

ankurp:
Adding a `with` option to `broadcasts_to` so that multiple broadcasts can be made for models where the streams are scoped with the associated model. For example in the example below we have a video call which has many to many relationship with users. Each user want to subscribe to video call updates that they are part of.

```ruby
class VideoCall < ApplicationRecord
  has_and_belongs_to_many :users
  broadcasts_to ->() { :video_calls }, with: :users
end
```

On the ERB template you would then just need to add the follow `turbo_stream_from ` tag

```erb
<%= turbo_stream_from [:video_calls, current_user] %>
```

Now any video calls that user is part of get added, updated or destroyed only those users get notified.

### DIFF
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -51,10 +51,32 @@ module Turbo::Broadcastable
     #     belongs_to :board
     #     broadcasts_to ->(message) { [ message.board, :messages ] }, inserts_by: :prepend, target: "board_messages"
     #   end
-    def broadcasts_to(stream, inserts_by: :append, target: broadcast_target_default)
-      after_create_commit  -> { broadcast_action_later_to stream.try(:call, self) || send(stream), action: inserts_by, target: target }
-      after_update_commit  -> { broadcast_replace_later_to stream.try(:call, self) || send(stream) }
-      after_destroy_commit -> { broadcast_remove_to stream.try(:call, self) || send(stream) }
+    #
+    # To broadcast to multiple unique channels for has_many associations
+    # Below is an example of a Video Call having many to many relationship to User model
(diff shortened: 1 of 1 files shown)


ankurp:
@dhh @kaspth Can you review whenever you get chance. I can make any modifications to PR as needed if you feel the approach here is not correct.

kaspth:
This is unfortunately the case where `has_and_belongs_to_many` masks a domain concept. So I'd extract the join model to a suitable domain concept:

```ruby
class VideoCall < ApplicationRecord
  has_many :participations
  has_many :users, through: :participations

  def participant_for(user)
    participations.find_by!(user: Current.user)
  end
end

class VideoCall::Participation < ApplicationRecord
  belongs_to :video_call
  belongs_to :user

  broadcasts
end
```

And then you can do `turbo_stream_from @video_call.participant_for(Current.user)`. It's slightly more code right now, but these type of extractions tend to pay off maintenance wise in the long term.

So I appreciate the suggestion, but I think we'll pass ✌️

ankurp:
@kaspth Your example is about a specific `@video_call` How can I subscribe to all video_calls for a user as new video calls can be created and a user may be added to it. In this case I dont have any handle to `@video_call` model.

kaspth:
Fair enough, I'd do `turbo_stream_from [ current_user, :video_calls ]` and then  `broadcasts_to ->(participation) { [ participation.user, :video_calls ] }` in `VideoCall::Participant` as the example in the documentation goes https://github.com/hotwired/turbo-rails/blob/cf5f3cb4d61b50f4c19b0b55aebba00c6f428e1a/app/models/concerns/turbo/broadcastable.rb#L52

PR #87, opened 2021-01-21
Remove unused variables

ar31an:
Remove unused **element**,  **dom_id** variables and use ternary condition.


kaspth:
Agree on reusing the target naming, thanks!

PR #90, opened 2021-01-22
Update tag_builder.rb - Fixes #102

### LINKED ISSUE #102: Turbo::Streams::TagBuilder sends wrong local variable to partial
When using namespaced models (like in an engine), TagBuilder sends the wrong local variable to partials.

**Example with a model `Spina::Image`**:
`<%= turbo_stream.update @image %>` assumes the correct partial using `to_partial_path`, but it uses `spina_image` (including the namespace) as the local variable, instead of `image`.

To use the correct variable we now have to explicitly write this:
`<%= turbo_stream.update @image, partial: @image.to_partial_path, locals: {image: @image} %>`

I'd rather write:
`<%= turbo_stream.update @image %>`

This is fixed in #90.

Bramjetten:
When using a model from an engine (using isolated_namespace), `singular` returns the model name prefixed with `myengine_`. This results in the wrong local variable. `param_key` returns the correct key without the engine prefix.

### DIFF
--- a/app/models/turbo/streams/tag_builder.rb
+++ b/app/models/turbo/streams/tag_builder.rb
@@ -122,7 +122,7 @@ class Turbo::Streams::TagBuilder
     def render_record(possible_record)
       if possible_record.respond_to?(:to_partial_path)
         record = possible_record
-        @view_context.render(partial: record.to_partial_path, locals: { record.model_name.param_key.to_sym => record }, formats: [ :html ])
+        @view_context.render(partial: record.to_partial_path, object: record, formats: :html)
       end
     end
 end
--- a/app/models/turbo/streams/tag_builder.rb
+++ b/app/models/turbo/streams/tag_builder.rb
@@ -122,7 +122,7 @@ class Turbo::Streams::TagBuilder
     def render_record(possible_record)
       if possible_record.respond_to?(:to_partial_path)
         record = possible_record
-        @view_context.render(partial: record.to_partial_path, locals: { record.model_name.singular.to_sym => record }, formats: [ :html ])
+        @view_context.render(partial: record.to_partial_path, locals: { record.model_name.param_key.to_sym => record }, formats: [ :html ])
       end
     end
 end
--- a/test/streams/streams_controller_test.rb
+++ b/test/streams/streams_controller_test.rb
@@ -37,4 +37,12 @@ STREAM
 STREAM
     end
   end
+  
+  test "render correct partial for namespaced models" do
+    get users_profile_path(id: 1), as: :turbo_stream
+    assert_equal <<-STREAM, @response.body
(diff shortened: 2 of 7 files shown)


app/models/turbo/streams/tag_builder.rb
```diff
@@ -121,7 +121,7 @@ def extract_target_name_from(target)
     def render_record(possible_record)
       if possible_record.respond_to?(:to_partial_path)
         record = possible_record
-        @view_context.render(partial: record.to_partial_path, locals: { record.model_name.singular.to_sym => record }, formats: [ :html ])
+        @view_context.render(partial: record.to_partial_path, locals: { record.model_name.param_key.to_sym => record }, formats: [ :html ])
```
kaspth, line 124:
```suggestion
        @view_context.render(partial: record.to_partial_path, object: record, formats: :html)
```

Can we just pass `object` instead and let the partial name dictate the variable name? Also don't think we need the wrapping array around the format.

kaspth (commented):
Nice!

app/models/turbo/streams/tag_builder.rb
Bramjetten: 💯, way better

kaspth:
Sweet, can we add a test to cover this behavior too?

Bramjetten:
Added a test with a namespaced `Users::Profile` model

test/streams/streams_controller_test.rb
```diff
@@ -37,4 +37,12 @@ class Turbo::StreamsControllerTest < ActionDispatch::IntegrationTest
 STREAM
     end
   end
+  
+  test "render correct partial for namespaced models" do
+    get users_profile_path(id: 1), as: :turbo_stream
+    assert_equal <<-STREAM, @response.body
```
seanpdoyle, line 43: Since this test inherits from ActionDispatch::IntegrationTest, we should have access to [assert_dom_equals](https://github.com/rails/rails-dom-testing#dom-assertions).

If that's the case, let's use it so that the test isn't dependent on the attribute order:

```ruby
test "render correct partial for namespaced models" do
  get users_profile_path(id: 1), as: :turbo_stream
  assert_dom_equal <<~HTML, @response.body
    <turbo-stream action="replace" target="users_profile_1"><template><p>User</p></template></turbo-stream>
    <turbo-stream action="update" target="users_profile_1"><template><p>User</p></template></turbo-stream>
  HTML
end
```

Bramjetten: I can change that, but I mimicked the other tests. Do these need to be changed to use `assert_dom_equal` as well?

seanpdoyle: I can touch up the others in a separate commit.

kaspth:
Sweet, thank you!

Issue #37, opened 2020-12-25
data-disable-with


jakehockey10:
Hello,

I thought I had read something about data-disable-with related to turbo-rails, but I can't find it in my history anywhere.  I notice that `data: { disable_with: '...' }` doesn't seem to work anymore when making the transition to turbo-rails from turbolinks and rails-ujs.

I wasn't quite sure what made data-disable-with work in the past, but I'm assuming it was rails-ujs since it doesn't seem to be related to turbolinks and the only other thing I removed to make way for turbo-rails was rails-ujs.

Are we planning on having data-disable-with functionality restored?  Or is this something we should implement ourselves from here on out?

I've really enjoyed learning about the new magic, thanks for all the efforts!

jakehockey10:
I found where I had read something that mentioned data-disable-with and turbo in the same conversation.  I wasn't able to find anything beyond these mentionings:

https://discourse.stimulusjs.org/t/stimulus-and-rails-ujs/587/10
https://discuss.hotwire.dev/t/are-transitions-and-animations-on-hotwire-roadmap/1547/3

dhh:
[Documented the compatibility with turbo-rails and UJS](9d53529bedd4de19dea8521d415f40685dfe7749). And we've also stopped taking UJS out when using the Webpack installer.

jakehockey10:
The link above didn't work for me, but here is where I found the documentation you mention: https://github.com/hotwired/turbo-rails#compatibility-with-rails-ujs

coorasse:
Also, seems duplicate of https://github.com/hotwired/turbo/issues/55

Issue #33, opened 2020-12-24
Links which specify a data-method have that method ignored


bhaibel:
Let's say I have the following in a template:

```
<%= link_to "Sign out", signout_path, method: :delete %>
```

When I click on it, I'll get a routing error that seems to reflect the link having been visited with a `GET` rather than a `DELETE` request. I believe that this means that `turbo-rails` is not picking up on the `data-method` in the rendered link.

dhh:
I thought we had already pulled this in from UJS, but guess not. @sstephenson thoughts on how to get this in?

Although an alternative would be to say that turning ahrefs into POSTs is a bit of an anti-pattern, especially for a11y reasons. Better to use button_to with a styling. I think that's why this didn't come up in HEY, since we've been removing links that mutate in part for a11y reasons.

But if that approach is followed, we clearly need to document it.

dhh:
@seanpdoyle Any thoughts on the a11y aspect?

bhaibel:
The a11y point is a really good one.

I think my preferred solution might be a two-part one:

bhaibel:
The a11y point is a really good one.

I still think we should port the old rails-ujs behavior in, at least for now (maybe it gets a deprecation warning slapped on it?) -- my instinct is that `data-method` links suddenly Stopping Working is going to be a point of confusion & frustration for folks trying to transition from UJS to Turbo.

From a developer UX perspective, my long-term ideal might be something that *warns* the developer they are doing a stupid a11y thing, but that might not be worth the extra kb over the wire.

seanpdoyle:
I agree, combining `link_to` and destructive HTTP Verbs is troublesome. From my perspective, the underlying mismatch between the semantics of a destructive action like an HTTP `DELETE` and an `<a>` element complicate the matter.

While the accessibility issues can be papered-over by declaring [role="link"](https://www.w3.org/TR/wai-aria-1.1/#link), it might still surprise users of assistive tech if clicking a link triggered anything but idempotent behavior.

> I still think we should port the old rails-ujs behavior in, at least for now (maybe it gets a deprecation warning slapped on it?) -- my instinct is that data-method links suddenly Stopping Working is going to be a point of confusion & frustration for folks trying to transition from UJS to Turbo.

[...]

PR #91, opened 2021-01-22
Document how Turbo works with each Rails UJS feature

ghiculescu:
After [this](https://github.com/hotwired/turbo/issues/119#issuecomment-765711374) back and forth I realised there's no docs on how the Rails UJS `method` feature works with Turbo. I added docs for this and also for the 3 other UJS features.

PR #93, opened 2021-01-25
Support for Unobtrusive JavaScript

seanpdoyle:
Closes https://github.com/hotwired/turbo-rails/issues/33.
Closes https://github.com/hotwired/turbo-rails/issues/37.

---

Re-create [hotwired/turbo#40][].

When loaded alongside [@rails/ujs][], start it via `Rails.start()`, and
then declare the [`turbo:`-prefixed events][turbo-events] that
correspond to their `ajax:`-prefixed counterparts.

Test suite changes
---

First, introduce System Test coverage to drive the Turbo and Rails
integration through a headless browser.

Next, change some dummy controller actions to accept a `?sleep=` query
parameter to delay their responses long enough to ensure that
`data-disabled-with` and `data-confirm` have a wide enough window of
opportunity for test coverage to exercise their behaviors.

[hotwired/turbo#40]: https://github.com/hotwired/turbo/pull/40
[@rails/ujs]: https://github.com/rails/rails/tree/v6.1.0/actionview/app/assets/javascripts
[turbo-events]: https://turbo.hotwire.dev/reference/events

### DIFF
--- a/test/dummy/app/views/messages/new.html.erb
+++ b/test/dummy/app/views/messages/new.html.erb
@@ -1,0 +1,22 @@
+<%= content_for :head do %>
+  <script type="module">
+    import Rails from "https://cdn.skypack.dev/@rails/ujs"
+
+    const { delegate, disableElement, enableElement } = Rails
+
+    delegate(document, Rails.linkDisableSelector,   "turbo:before-cache", enableElement)
+    delegate(document, Rails.buttonDisableSelector, "turbo:before-cache", enableElement)
+    delegate(document, Rails.buttonDisableSelector, "turbo:submit-end", enableElement)
+
+    delegate(document, Rails.formSubmitSelector, "turbo:submit-start", disableElement)
+    delegate(document, Rails.formSubmitSelector, "turbo:submit-end", enableElement)
+    delegate(document, Rails.formSubmitSelector, "turbo:before-cache", enableElement)
+  </script>
+<% end %>
+
+<a href="<%= new_message_path(sleep: 2) %>" data-remote="true" data-disable-with="Reloading">Reload</a>
+
+<%= form_with url: messages_path(sleep: 2), data: { remote: true } do %>
+  <button data-disable-with="Submitting">Submit</button>
+  <button data-confirm="Are you sure?">Confirm</button>
+<% end %>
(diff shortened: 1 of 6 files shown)


test/dummy/app/views/messages/new.html.erb
```diff
@@ -0,0 +1,22 @@
+<%= content_for :head do %>
+  <script type="module">
+    import Rails from "https://cdn.skypack.dev/@rails/ujs"
+
+    const { delegate, disableElement, enableElement } = Rails
```
seanpdoyle, line 5: Similar to the question raised on hotwired/turbo#40 (https://github.com/hotwired/turbo/pull/40#discussion_r548945358), it isn't clear on where this code should live.

One option is to declare it as an engine-provided HTML partial declaring the `<script type="module">` in the same style as this test currently does. Consumers could call it like any other Turbo-provided helpers. Unfortunately, there wouldn't be a seam for them to provide their applications already loaded `Rails` from `@rails/ujs`.

Another alternative would be to declare a separate `@turbo-rails/ujs` module for consumers to import and load (calling something like `startUJS(Rails)` so that they can provide their own already started `Rails`. Unfortunately, during my initial attempts at that, it wasn't trivial to import _that_ module into this engine's test suite's `test/dummy/packs/application.js`. 

If there's a clear path forward, I'll re-structure this block in any way that suits us. In its current form, it at least provides confidence that a polyfill is possible and able to be covered by automated tests.

seanpdoyle: Once we decide on how consumers can consume it, I'll follow up with some documentation changes (to the README or elsewhere).

dhh:
This is interesting, but I'd almost rather just find a native way to deal with the functionality in Hotwire. Maybe as Stimulus standard controllers? Or, if we're going to ship a shim, have turbo-rails simply contain the functionality itself. It's very little code.

ghiculescu:
> This is interesting, but I'd almost rather just find a native way to deal with the functionality in Hotwire.

@dhh @seanpdoyle As a start, what do you think about handling it through docs? https://github.com/hotwired/turbo-rails/pull/91

(Stimulus standard controllers to replace the current UJs behaviors would be nice too. Which project would that live in - presumably Stimulus?)

PR #98, opened 2021-01-28
Ensure html format when rendering turbo_stream responses. Fixes #97

### LINKED ISSUE #97: Explicitly rendering a different turbo stream view breaks format context
I have a repo in "reproduction steps" below that have a simplified example of an issue I ran into with making a turbo stream POST request and explicitly rendering a turbo stream view that has an html partial in it.

I have a turbo frame, called "form-area". I have a link (let's say to a new post) that points to the frame. The new action explicitly renders a `new.turbo_stream.erb` view in it and replaces the frame as expected. That part works just fine.

```html
<!-- Main page -->
<a href="posts/new" data-turbo-frame="form-area">New Post</a>

<turbo-frame id="form-area"></turbo-frame>
```

```ruby
# new.turbo_stream.erb
<%= turbo_stream.update "form-area" do %>
  <div>Special markup I'd like wrapped around the form in a frame</div>
  <%= render 'form', post: @post %>
<% end %>
```

On the create action when I submit the form, a validation error is triggered. I choose in my respond_to block for format.turbo_stream to render :new  which means "render the new view for turbo streams. That part starts to work fine.

```ruby
# posts_controller.rb

  def create
    @post = Post.new(post_params)

    respond_to do |format|
      if @post.save
        format.html { redirect_to @post, notice: "Post was successfully created." }
        format.turbo_stream { render :new }
        format.json { render :show, status: :created, location: @post }
      else
        format.html { render :new, status: :unprocessable_entity }
        format.turbo_stream { render :new }
        format.json { render json: @post.errors, status: :unprocessable_entity }
      end
    end
  end
```

HOWEVER, when the turbo stream view hits the partial (remember, I am explicitly telling the controller to render this view, I get an UnknownFormat error as turbo is not (I assume?) under the covers switching the view context to HTML to look for html partials instead of more turbo stream partials.

```
ActionView::Template::Error (Missing partial posts/_form, application/_form with {:locale=>[:en], :formats=>[:turbo_stream], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :jbuilder]}. Searched in:
  * "/.../possible_turbo_issue/app/views"
  * "/.../.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/actiontext-6.1.1/app/views"
  * "/.../.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/actionmailbox-6.1.1/app/views"
):
    1: <%= turbo_stream.update "form-area" do %>
    2:   <div>Special markup I'd like wrapped around the form in a frame</div>
    3:   <%= render 'form', post: @post %>
    4: <% end %>

app/views/posts/new.turbo_stream.erb:3
app/views/posts/new.turbo_stream.erb:1
app/controllers/posts_controller.rb:32:in `block (2 levels) in create'
app/controllers/posts_controller.rb:26:in `create'
```

Reproduction Steps:
1. Clone repo: https://github.com/t27duck/possible_turbo_issue
2. Run `bundle install`
3. Run migrations
4. Start `rails s`
5. Go to localhost 3000
6. Click "New Post"
7. Don't fill out the form and click submit
8. See the exception in the Rails log

excid3:
This PR fixes #97 by ensuring the `html` format is included whenever a turbo_stream tag is rendered.

**Example**

Using format.turbo_stream changes `formats` from `[:turbo_stream, :html]` to just `[:turbo_stream]` causing a normal render to fail inside the turbo_stream block.

```ruby
format.turbo_stream { render :new }
```

```erb
<%# new.turbo_stream.erb %>

<%= turbo_stream.update "form-area" do %>
  <%= render 'form', post: @post %>
<% end %>
```

Results in 

```
ActionView::Template::Error (Missing partial posts/_form, application/_form with {:locale=>[:en], :formats=>[:turbo_stream], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :jbuilder]}.
```

This change ensures that `:html` is in the formats list when `turbo_stream` is called.

cc @kaspth @t27duck


excid3:
Alternative could be to modify the formats temporarily and restore them around the block render here:

```ruby
when block_given?
  # save formats, add :html to formats 
  turbo_stream_action_tag name, target: target_name, template: @view_context.capture(&block)
  # Restore 
```

However, I figured that anytime you use `turbo_stream` you're going to want html in the formats list anyways.

app/models/turbo/streams/tag_builder.rb
```diff
@@ -27,6 +27,7 @@ class Turbo::Streams::TagBuilder
 
   def initialize(view_context)
     @view_context = view_context
+    @view_context.formats |= [:html]
```
kaspth, line 30: Will this also change things for the controller level tag builder? I think it's best we keep the no-fallback there, no? Unless well, I don't really remember why that's there.

We could add a `fallback_format:` to the builder intializer and then only include that in the view tag builder.

kaspth (commented):
Nice and simple!

app/models/turbo/streams/tag_builder.rb
excid3: The controller method just instantiates this class, so it would affect that too. 

I think it makes sense to have `format.turbo_stream` either call `turbo_stream` or render a `turbo_stream.erb` template. You wouldn't just not specify it if you wanted to use `html.erb` response. 👍

t27duck:
I can confirm using this branch against my test repo from the issue that this does the job.

app/models/turbo/streams/tag_builder.rb
t27duck: Having the fallback keeps this inline with having an action with implicit rendering...

```ruby
def new
  @post = Post.new
end
```

Where if a TURBO_STREAM request comes in, the `format` is already set to `[:turbo_stream, :html]` so it would not only look for a turbo stream view first but also that format would be passed through to the view if it has any partcials.

kaspth: @excide3 Yeah, I get that point. Something irks me about altering the formats from within a `TagBuilder` though, especially since template rendering has been a rife source of security vulnerabilities in the past. So we do have to be a little careful here.

I'm leaning towards the block version you posted earlier 👍 @dhh may have a different take though 😄

dhh:
Very simple! Let's add some tests.

excid3:
Added a test for this. 👍

app/models/turbo/streams/tag_builder.rb
excid3: This works if we want to go that route:

```ruby
    when block_given?
      with_html_format do
        turbo_stream_action_tag name, target: target_name, template: @view_context.capture(&block)
      end
```
```ruby
    def with_html_format
      previous_formats = @view_context.formats
      @view_context.formats |= [:html]
      result = yield
      @view_context.formats = previous_formats
      result
    end
```

kaspth: We'd need an `ensure` for the reset to guard for exceptions and then we could just remove the `result` variable, but yes, that would be the other way. Let's go with the mutation for now and see how that goes!

kaspth:
🙌

app/models/turbo/streams/tag_builder.rb
excid3: Ah yes, of course! As I was writing that I was like I know I'm missing something. 😜

PR #105, opened 2021-01-30
Improve implicit view path handling in streams tag builder

agrobbin:
If you are in a controller that has added something to the view path, like this:

```ruby
class Admin::BaseController < ApplicationController
  prepend_view_path 'app/views/admin'
end

class Admin::CompaniesController < Admin::BaseController
  def show
  end
end
```

The default rendering within `Streams::TagBuilder` does not do the right partial path lookup within the view path, due to it explicitly passing `partial: record.to_partial_path` to `render`. It also happens to cause some interesting I18n lookup issues, due to the incorrect virtual path.

By passing `record` itself, we get to take advantage of Action View's built-in partial path generation and virtual path logic, which accounts for any custom view paths.

It also happens to be even simpler to implement!

This is essentially an extension of the work already done in #90.

### DIFF
--- a/app/models/turbo/streams/tag_builder.rb
+++ b/app/models/turbo/streams/tag_builder.rb
@@ -122,7 +122,7 @@ class Turbo::Streams::TagBuilder
     def render_record(possible_record)
       if possible_record.respond_to?(:to_partial_path)
         record = possible_record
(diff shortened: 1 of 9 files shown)


kaspth:
Perfect! 🙌

agrobbin:
Wow, so fast, thanks @kaspth!

PR #108, opened 2021-02-01
Add System Test coverage for broadcasting

seanpdoyle:
Introduce System Test coverage to drive the Turbo and Rails integration
through a headless browser.

Turbo Stream broadcasting relies on both server-side behavior and
client-side behavior. ActionDispatch::Integration tests are useful for
covering particular edge cases, but in-browser System Tests enable tests
to exercise the entire spectrum.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,3 +4,11 @@ gemspec
 
 gem 'rake'
 gem 'byebug'
+gem 'puma'
+
+group :test do
+  gem 'capybara'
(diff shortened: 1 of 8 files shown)


kaspth:
@seanpdoyle can you take a look at the test failure?

seanpdoyle:
@kaspth I've added some Capybara presence assertions to better synchronize the system tests. They should pass consistently now!

PR #109, opened 2021-02-01
fix namescaped model naming when broadcast to partial

### LINKED ISSUE #103: Inconsistent local naming with namespace model
Given a normal model and a namespaced one:

```ruby
class Car < ActiveRecord::Base
end

@car = Car.first
@car.to_partial_path # => "cars/car"
@car.model_name.singular # => "car"
```

You can render the partial with `render @car` which will set `car` as a local variable. All expected.

With a namescaped model...

```ruby
class Space::Car < ActiveRecord::Base
end

@car = Space::Car.first
@car.to_partial_path # => "space/cars/car"
@car.model_name.singular # => "space_car"
```

When rendering in the view with the namespaced model using `render @car`, it will set `car` as a local variable as that is the name of the partial. However, when rendering via the `broadcastable` concern, it doesn't use the partial name to set the local variable name but [the singular version of the model name](https://github.com/hotwired/turbo-rails/blob/main/app/models/concerns/turbo/broadcastable.rb#L236). The only way to get this is work both ways is to manually override the `partial` and `locals` for each broadcast.

I'm not sure what the best path forward is but it seems that the naming of the local variable should be consistent whether rendering from a view or from a broadcast.

ryanwood:
Fixes #103. I *think* this should work but would like some other eyes to see if this is the best way to fix it. I modeled it after #90. Once issue is that I'm getting a "\n" in my rendered template and I'm not sure why.

Happy to make any adjustments. Thanks.

### DIFF
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -233,7 +233,7 @@ module Turbo::Broadcastable
 
     def broadcast_rendering_with_defaults(options)
       options.tap do |o|
-        o[:locals]    = (o[:locals] || {}).reverse_merge!(model_name.singular.to_sym => self)
+        o[:object] ||= self
         o[:partial] ||= to_partial_path
       end
     end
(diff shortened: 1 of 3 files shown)


kaspth:
Sweet, looks legit 🙌

ryanwood:
@kaspth Could you release a new version with this fix?

PR #110, opened 2021-02-03
Turbo-rails should be more specific about its dependencies

atavistock:
I typically exclude parts of rails that I don't actually use.  Some are just to avoid bloat like actionmailbox, others impose behaviors that I may not actually want like sprockets.   

This reduces the runtime dependencies down to what is actually used by turbo-rails.  Tests still run using activerecord, sprockets, etc as dev-dependencies.


dhh:
Closed via #283

PR #112, opened 2021-02-05
Update method names in Broadcastable "same as" documentation comments

mikej:
Assuming I've correctly understood the intention of the various methods in the `Broadcastable` concern it looks like some of the documentation comments weren't quite right.

This PR updates these and I think they're all correct now.

Let me know if I've misunderstood anything here. Thanks.


mikej:
Thanks @dhh 😄

PR #113, opened 2021-02-05
Allow rendering to be passed so that we can change the partial and object when broadcasting

ankurp:
In the broadcasts_to method we cannot change the partial if we want to render the model differently on certain pages.

Example if I need to specify a partial I need to write it out like this
```ruby
  after_create_commit ->(video_call) { broadcast_append_to [:video_calls], partial: "video_calls/video_call_row" }
  after_update_commit ->(video_call) { broadcast_replace_to [:video_calls], partial: "video_calls/video_call_row" }
  after_destroy_commit ->(video_call) { broadcast_remove_to [:video_calls] }
```

Proposal is to allow partial to be passed in broadcasts_to

```ruby
broadcasts_to ->(video_call) { [:video_calls] }, target: "video_calls", partial: "video_calls/video_call_row"
```


dhh:
Can you add tests for this?

ankurp:
@dhh @kaspth can one of you help me edit the test. I am not able to figure out how to invoked the create callback so that it performs the broadcast later which we want to check if triggered.

dhh:
After #115 you should be able to use some of these new db-backed fixtures to test.

app/models/concerns/turbo/broadcastable.rb
```diff
@@ -51,10 +51,10 @@ module ClassMethods
...
     #     broadcasts_to ->(message) { [ message.board, :messages ] }, inserts_by: :prepend, target: "board_messages"
     #   end
-    def broadcasts_to(stream, inserts_by: :append, target: broadcast_target_default)
-      after_create_commit  -> { broadcast_action_later_to stream.try(:call, self) || send(stream), action: inserts_by, target: target }
-      after_update_commit  -> { broadcast_replace_later_to stream.try(:call, self) || send(stream) }
-      after_destroy_commit -> { broadcast_remove_to stream.try(:call, self) || send(stream) }
+    def broadcasts_to(stream, inserts_by: :append, target: broadcast_target_default, **rendering)
+      after_create_commit  -> { broadcast_action_later_to stream.try(:call, self) || send(stream), action: inserts_by, target: target, **rendering }
+      after_update_commit  -> { broadcast_replace_later_to stream.try(:call, self) || send(stream), **rendering }
+      after_destroy_commit -> { broadcast_remove_to stream.try(:call, self) || send(stream), **rendering }
```
genezys, line 57: I suggest removing the rendering options for `broadcast_remove_to` that does not accept them. I don't see how that would make sense.

PR #115, opened 2021-02-07
Enable broadcastable target to be namespaced by parent

### LINKED ISSUE #114: Overriding the broadcasting target for nested models
Hello!

## Problem description

I encountered an issue today while learning Hotwire by working on a toy app with nested models. I am building a quote editor, where a quote has many line_item_dates and dates has many line_items. The models look like this:

```rb
# models/quote.rb

class Quote < ApplicationRecord
  has_many :line_item_dates, dependent: :destroy
  
  broadcasts_to ->(quote) { :quotes }, inserts_by: :prepend
end
```

```rb
# models/line_item_date.rb

class LineItemDate < ApplicationRecord
  belongs_to :quote
  has_many :line_items, dependent: :destroy
  
  validates :date, presence: true, uniqueness: { scope: :quote }
  
  broadcasts_to ->(line_item_date) { [line_item_date.quote, :line_item_dates] }, inserts_by: :prepend
end
```

```rb
class LineItem < ApplicationRecord
  belongs_to :line_item_date

  validates :name, presence: true
  validates :quantity, :price, presence: true, numericality: { only_integer: true }
  
  # This is where I have the problem, as there are multiple "line_items" targets on the page
  # I would like to "namespace" the "line_items" target by the dom_id of the line_item_date.
  # For example instead of having multiple "line_items" targets, I would like to have one 
  # "line_item_date_1_line_items"  target and one "line_item_date_2_line_items" target.
  broadcasts_to ->(item) { [item.line_item_date, :line_items] }
end
```

Here is a picture of my toy project to make it more clear (I can also share the code if you want). As you can see, there are two line_item_dates and two line_items. If I create a new line_item, for the second line_item_date, I want it to appear in the second line_item_date card. In other words, I want it to be appended to the turbo_frame with the target "line_item_date_2_line_item_dates"

Screenshot 2021-02-07 at 15 14 56

## Solution

It should be possible to "namespace" the targets to avoid having multiple targets of the same name on the same page. The class method `.broadcast_target_default` in Broadcastable is not specific enough to identify targets when there are multiple targets with the same name on the page (in the case of a quote editor for example).

I would *love* (😍) to contribute on this issue, I will submit a pull request this afternoon!

This is my (almost) first contribution to Rails as I opened my first pull request in the `rails` main repo a few days ago so I hope I made things right. Please tell me if my message above is not clear enough!

alexandreruban:
Closes #114 

The class method `.broadcast_target_default` now delegates to the instance method `#broadcast_target_default`. It is still equal to `model_name.plural` by default but can now easily be overridden at the instance level. This is especially useful if you want to “namespace” targets by the parent name and id. 

Consider these two models:

```rb
  has_many :comments 

  # Default target is "messages"
  broadcasts_to ->(message) { :messages }
end
```

```rb 
class Comment < ApplicationRecord
  belongs_to :message

  # Default target is "message_#{message.id}_comments"
  broadcasts_to ->(comment) { [comment.message, :comments] }

  private

  def broadcast_target_default
    "message_#{message.id}_comments"
  end
end
```

Without these changes, if you display mutiple messages and comments on the same page, you will end up with more than one "comments" target. Instead, you want to namespace them by the dom_id of the parent message.

Please tell me if this is OK or if it requires changes! :)


app/models/concerns/turbo/broadcastable.rb
```diff
@@ -52,21 +52,21 @@ module ClassMethods
...
-      after_create_commit  -> { broadcast_action_later action: inserts_by, target: target }
+      after_create_commit  -> { broadcast_action_later action: inserts_by, target: target.send(:call, self) || target }
       after_update_commit  -> { broadcast_replace_later }
       after_destroy_commit -> { broadcast_remove }
     end
 
     # All default targets will use the return of this method. Overwrite if you want something else than <tt>model_name.plural</tt>.
     def broadcast_target_default
-      model_name.plural
+      ->(broadcastable) { broacastable.send(:broadcast_target_default) }
```
alexandreruban, line 69: Does it make sense to make instance method `#broadcast_target_default` a public method?

blaknite:
Hey @alexandreruban I'd love to see this functionality exist. I hit a bug in my app just now that needed targets scope to the parent model. I've monkey patched `AppicationRecord` with what you've provided here and it works.

As an FYI there's more calls to `broadcast_target_default` within the concern you've edited that will likely explode with the change as is. If I get some time in the next few days I'll flesh those out.

dhh:
Don't like the idea of an instance method for broadcast_target_default, but do like the idea that you can pass a proc as a target. Let's just slim it down to just that.

alexandreruban:
@dhh thank you for the feedback!

I reverted my changes to only keep the part that enables to pass a proc as a target. Please tell me if it's OK for you now!

dhh:
Looks good, just need to add some test coverage 👍

alexandreruban:
I added tests for both `Turbo::Broadcastable.broadcasts_to` and `Turbo::Broadcastable.broadcasts` class methods. I added a database in order to test the override with the proc and the whole behaviour with the `after_(create|update|destroy)_commit` callbacks. This is why I created two ActiveRecord models, one with `broadcasts_to` and one with `broadcasts`.

Please tell me if it works for you and thanks again for your feedback! @dhh

PR #124, opened 2021-02-13
Webpacker install: Allow require("@rails/activestorage")

robbevp:
Changes the webpacker install to find either `import ActiveStorage` or `require("@rails/activestorage")`

`require("@rails/activestorage")` must have been the Rails default at some point, since I had it in my repo since the initial commit.


dixpac:
I think this is good. Idea behind using just `import` was that latest rails version ships with `import`, but this will improve resiliency of the generator since older version used `require` 👍🏼

PR #131, opened 2021-02-22
Clear a target's content from the DOM

ayrton:
Introduce a little helper to clear the content of a target in the DOM. This is useful eg when you are closing a modal

```rb
def create
  # ...

  respond_to do |format|
    format.turbo_stream
    format.html { redirect_to user_root_path }
  end
end
```

__Before:__

```
# create.turbo_stream.erb
<%= turbo_stream.replace :modal, "" %>
```

__After:__

```
# create.turbo_stream.erb
<%= turbo_stream.clear :modal %>
```

### DIFF
--- a/app/models/turbo/streams/tag_builder.rb
+++ b/app/models/turbo/streams/tag_builder.rb
@@ -40,6 +40,16 @@ class Turbo::Streams::TagBuilder
     action :remove, target, allow_inferred_rendering: false
   end
 
+  # Clear the <tt>target</tt> in the dom. The target can either be a dom id string or an object that responds to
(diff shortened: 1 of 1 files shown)


dhh:
Feels like this should live in Turbo, and then we an call it from turbo-rails. Want to take a stab at adding it as a basic command first that we can then invoke?

ayrton:
> Feels like this should live in Turbo, and then we an call it from turbo-rails. Want to take a stab at adding it as a basic command first that we can then invoke?

Opened PR https://github.com/hotwired/turbo/pull/241. Will follow up in this repo once merged in

PR #141, opened 2021-03-04
Add .ruby-version and .ruby-gemset to assist developers who use RVM.

gregschmit:
Not sure if you guys are interested in this, but it would just make development for people who use RVM easier, since RVM would auto-switch to Ruby 2.7 and a `turbo-rails` gemset so they can `bundle install` and `rake test` quickly to get started.


dhh:
The default for Rails is to use rbenv, so we'll stick with that.

gregschmit:
I think rbenv respects the `.ruby-version` file, so not sure why you wouldn't have that in this repo.

PR #142, opened 2021-03-04
Introduce disableDriveByDefault method to allow users to disable Drive by default.

### LINKED ISSUE #139: Disable Turbo Drive by default
In my project, I import Turbo (via Webpacker) like this:

    import "@hotwired/turbo-rails";

For reasons (it breaks my existing Vue.js stuff), I want to disable Turbo Drive. So I added `data-turbo="false"` to my `<html>` tag in my layout. But for another project, that's kinda cumbersome (and in my scenario, there are customer-written layouts and I don't want them to have to worry about stuff like that).

Is there a way to disable Turbo Drive entirely? Either via a Rails initializer configuration, or by replacing the import with a set of import statements that import all the components separately? Or perhaps a variable I can set or a JS statement I can make before/after Turbo is imported?

gregschmit:
It can then be enabled on a per-element basis with `data-turbo="true"`. I also added some documentation to the README.md.

Usage is provided in the edit of README.

Fixes #139

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,15 @@ During rendering, Turbo replaces the current `<body>` element outright and merge
 
 Whereas Turbolinks previously just dealt with links, Turbo can now also process form submissions and responses. This means the entire flow in the web application is wrapped into Turbo, making all the parts fast. No more need for `data-remote=true`.
 
+Turbo Drive can be disabled on a per-element basis by annotating the element or any of its ancestors with `data-turbo="false"`. If you want Turbo Drive to be disabled by default, then you can adjust your import like this:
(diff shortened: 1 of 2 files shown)


gregschmit:
Shit... is this kinda change supposed to go to https://github.com/hotwired/turbo?

gregschmit:
Maybe if https://github.com/hotwired/turbo is ok with merging the PR I submitted there, I can remove the JS code here but still leave the change in README, so the feature would be clear to people who visit this repo?

dhh:
Near needs to happen in turbo.

PR #143, opened 2021-03-07
Discard broadcast jobs on deserialization error

robinjam:
`Turbo::Streams::ActionBroadcastJob` and `Turbo::Streams::BroadcastJob` can fail if they reference AR records that are deleted after the job has been enqueued but before it has run.

In this case it probably makes sense to discard the job rather than retrying it indefinitely.

### DIFF
--- a/app/jobs/turbo/streams/broadcast_job.rb
+++ b/app/jobs/turbo/streams/broadcast_job.rb
@@ -1,6 +1,8 @@
 # The job that powers the <tt>broadcast_render_later_to</tt> available in <tt>Turbo::Streams::Broadcasts</tt> for rendering
 # turbo stream templates.
 class Turbo::Streams::BroadcastJob < ActiveJob::Base
+  discard_on ActiveJob::DeserializationError
+  
   def perform(stream, **rendering)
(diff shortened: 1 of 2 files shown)


app/jobs/turbo/streams/broadcast_job.rb
```diff
@@ -1,6 +1,8 @@
 # The job that powers the <tt>broadcast_render_later_to</tt> available in <tt>Turbo::Streams::Broadcasts</tt> for rendering
 # turbo stream templates.
 class Turbo::Streams::BroadcastJob < ActiveJob::Base
+  discard_on ActiveJob::DeserializationError
```
seanpdoyle, line 4: Could we add tests to cover this behavior?

Maybe something that creates a record, enqueues a job, [performs the jobs](https://edgeapi.rubyonrails.org/classes/ActiveJob/TestHelper.html#method-i-perform_enqueued_jobs), then asserts that [the job was only run once](https://edgeapi.rubyonrails.org/classes/ActiveJob/TestHelper.html#method-i-assert_performed_jobs) and that another is [not enqueued](https://edgeapi.rubyonrails.org/classes/ActiveJob/TestHelper.html#method-i-assert_enqueued_jobs).

PR #153, opened 2021-03-31
[RFC] Add remove_if_present to append and prepend (avoids double-insert)

jeromecornet:
This is a request for comments/opinions.

In the demo app, @dhh specifically commented out the turbo_stream response to the action to avoid double rendering the message: https://github.com/hotwired/hotwire-rails-demo-chat/blob/0fb439d756ac59a1dd41c978973cc04d0517edd4/app/views/messages/create.turbo_stream.erb#L1

The problem with this approach is that
- it's not resilient to broken action cable connection (unlike a turbo_stream response to the action)
- there is a delay from the sender of the message (the background job needs to be executed, then sent through action cable)

One potential solution to this is to perform the action twice (once in the turbo_stream response, one in the broadcast job) but make the action idempotent.

What do you think about augmenting append and prepend with an 'upsert'-like function whereby:
- if an dom id is present, the rendered template will replace the matching element
- otherwise, the append/prepend method will add the contents of the rendered template to the target container (as per the existing)

This way, the first response (likely the turbo_stream) will append the message to the container, and the second action (from the broadcast job) will replace the message in place.

I took a crack at it in this PR by passing a `replace_if_present:` option to the append and prepend calls (and related methods).

On the plus side, it makes the client code fairly easy (add `replace_if_present` to both the Turbo::Streams::ActionBroadcastJob and the turbo_stream response) as per this commit: https://github.com/jeromecornet/hotwire-rails-demo-chat/commit/696e1c91b8eef2d64d54d387bf1462526aca5fd8

But the code looks not very pretty, as it introduces an append & prepend -specific option in many methods throughout the code.

What do you think of this approach ? Do you think it would be worth pursuing ? Any ideas on solving it better ?

If you think it's a good idea, I will clean this up before submitting a proper PR.

### DIFF
--- a/app/assets/javascripts/turbo.js
+++ b/app/assets/javascripts/turbo.js
@@ -1191,11 +1191,21 @@ function activateElement(element) {
 const StreamActions = {
   append() {
     var _a;
-    (_a = this.targetElement) === null || _a === void 0 ? void 0 : _a.append(this.templateContent);
+    if (!this.replaceIfPresent || ((_a = this.replaceIfPresentElement) === null || _a === void 0)) {
+      (_a = this.targetElement) === null || _a === void 0 ? void 0 : _a.append(this.templateContent);
+    }
+    else {
+      _a.replaceWith(this.templateContent)
+    }
   },
   prepend() {
     var _a;
-    (_a = this.targetElement) === null || _a === void 0 ? void 0 : _a.prepend(this.templateContent);
+    if (!this.replaceIfPresent || ((_a = this.replaceIfPresentElement) === null || _a === void 0)) {
+      (_a = this.targetElement) === null || _a === void 0 ? void 0 : _a.prepend(this.templateContent);
+    }
+    else {
+      _a.replaceWith(this.templateContent)
+    }
   },
   remove() {
     var _a;
@@ -1254,6 +1264,13 @@ class StreamElement extends HTMLElement {
     }
     this.raise("target attribute is missing");
   }
+  get replaceIfPresentElement(){
+    var _a;
+    if (this.replaceIfPresent) {
+      return (_a = this.ownerDocument) === null || _a === void 0 ? void 0 : _a.getElementById(this.replaceIfPresent);
+    }
+    this.raise("target attribute is missing");
+  }
   get templateContent() {
     return this.templateElement.content;
   }
@@ -1268,6 +1285,9 @@ class StreamElement extends HTMLElement {
   }
   get target() {
     return this.getAttribute("target");
+  }
+  get replaceIfPresent(){
+    return this.getAttribute("replace_if_present");
   }
   raise(message) {
     throw new Error(`${this.description}: ${message}`);
--- a/app/assets/javascripts/turbo.js
+++ b/app/assets/javascripts/turbo.js
@@ -1191,21 +1191,11 @@ function activateElement(element) {
 const StreamActions = {
   append() {
     var _a;
-    if (!this.replaceIfPresent || ((_a = this.replaceIfPresentElement) === null || _a === void 0)) {
-      (_a = this.targetElement) === null || _a === void 0 ? void 0 : _a.append(this.templateContent);
-    }
-    else {
-      _a.replaceWith(this.templateContent)
-    }
+    (_a = this.targetElement) === null || _a === void 0 ? void 0 : _a.append(this.templateContent);
   },
   prepend() {
     var _a;
-    if (!this.replaceIfPresent || ((_a = this.replaceIfPresentElement) === null || _a === void 0)) {
-      (_a = this.targetElement) === null || _a === void 0 ? void 0 : _a.prepend(this.templateContent);
-    }
-    else {
-      _a.replaceWith(this.templateContent)
-    }
+    (_a = this.targetElement) === null || _a === void 0 ? void 0 : _a.prepend(this.templateContent);
   },
   remove() {
     var _a;
@@ -1264,13 +1254,6 @@ class StreamElement extends HTMLElement {
     }
     this.raise("target attribute is missing");
   }
-  get replaceIfPresentElement(){
-    var _a;
-    if (this.replaceIfPresent) {
-      return (_a = this.ownerDocument) === null || _a === void 0 ? void 0 : _a.getElementById(this.replaceIfPresent);
-    }
-    this.raise("target attribute is missing");
-  }
   get templateContent() {
     return this.templateElement.content;
   }
@@ -1285,9 +1268,6 @@ class StreamElement extends HTMLElement {
   }
   get target() {
     return this.getAttribute("target");
-  }
-  get replaceIfPresent(){
-    return this.getAttribute("replace_if_present");
   }
   raise(message) {
     throw new Error(`${this.description}: ${message}`);
--- a/app/helpers/turbo/streams/action_helper.rb
+++ b/app/helpers/turbo/streams/action_helper.rb
@@ -12,7 +12,8 @@ module Turbo::Streams::ActionHelper
 
     if replace_if_present
       replacement = convert_to_turbo_stream_dom_id(replace_if_present)
-      %(<turbo-stream action="#{action}" target="#{target}" replace_if_present="#{replacement}">#{template}</turbo-stream>).html_safe
+      %(<turbo-stream action="remove" target="#{replacement}"></turbo-stream>
+        <turbo-stream action="#{action}" target="#{target}">#{template}</turbo-stream>).html_safe
     else
       %(<turbo-stream action="#{action}" target="#{target}">#{template}</turbo-stream>).html_safe
     end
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/turbo.js
```diff
@@ -1269,6 +1286,9 @@ class StreamElement extends HTMLElement {
   get target() {
     return this.getAttribute("target");
   }
+  get replaceIfPresent(){
```
seanpdoyle, line 1289: The codebase's copy of Turbo comes from [hotwired/turbo](https://github.com/hotwired/turbo/), so any JavaScript changes should be proposed there.

With that being said, this PR seems to share some overlap with https://github.com/hotwired/turbo/pull/191.

The [guidance provided in that PR](https://github.com/hotwired/turbo/pull/191#pullrequestreview-602190778) was to handle it as part of the `broadcast_{append,prepend}_to` suite of helpers. 

You're proposing something similar here. Instead of making `replace_if_present:` an option, have you tried combining `broadcast_remove` with the append and prepend broadcasts?

jeromecornet: Sorry I didn't look into the turbo repo :-/ 

It looks like what you did in hotwired/turbo#191 would do this, whereas my version was opt-in.
But indeed, the end effect would be the same as doing a remove/append in the templates.

I will close this, thanks for the heads up !

seanpdoyle: @jeromecornet No worries! I don't think https://github.com/hotwired/turbo/pull/191 is viable, but I think your proposed changes are. Would you be interested in transforming your proposal to incorporate the remove+append style of broadcast?

jeromecornet: ok, reopening to see where it can go.... 

I reverted the changes to turbo.js and replaced the turbo tag generation with a remove + append in https://github.com/hotwired/turbo-rails/pull/153/commits/e8aa95f9165636fd69778905f352e89f9766ee2e

jeromecornet:
@seanpdoyle I took another pass at it following  https://github.com/hotwired/turbo-rails/pull/153#discussion_r605187265 

Given the implementation, I renamed the option to `remove_if_present` throughout, and got rid of the changes to turbo.js so it's all local to this repo now. 

The demo chat app looks like this after this change:
```ruby
# app/models/message.rb
  belongs_to :room
  broadcasts_to :room, remove_if_present: :itself
end

# app/view/messages/create.turbo_stream.erb
<%= turbo_stream.append :messages, @message, remove_if_present: @message %>
```

Thoughts ?

app/helpers/turbo/streams/action_helper.rb
```diff
@@ -6,11 +6,20 @@ module Turbo::Streams::ActionHelper
...
-  def turbo_stream_action_tag(action, target:, template: nil)
+  #
+  # If remove_if_present is specified, the helper will generate 2 tags (remove and the original action)
+  #
+  def turbo_stream_action_tag(action, target:, template: nil, remove_if_present: nil)
     target   = convert_to_turbo_stream_dom_id(target)
     template = action.to_sym == :remove ? "" : "<template>#{template}</template>"
 
-    %(<turbo-stream action="#{action}" target="#{target}">#{template}</turbo-stream>).html_safe
+    if remove_if_present
```
seanpdoyle, line 16: As an alternative to accepting and passing around an `remove_if_present:` option, would it be safe to infer the value from `action.in? %[ append prepend ]` within the `ActionHelper#turbo_stream_action_tag` method?

That way, callers are not responsible for making the decision themselves, and `prepend` and `append` actions are always going to remove counterparts with matching `[id]` attributes.

jeromecornet: Right... the issue is how to recover the dom id of the rendered element itself (the `target` parameter  is the id of the container, not the element).
If it's not passed through the call chain, I'm not sure how to recover it. Are you thinking of parsing the template to extract the id of the element ?

seanpdoyle: Ah, that's a great point. Maybe the logic belongs higher in the call stack, closer to the code that maps the ActiveModel instance to the broadcast operation.

app/models/concerns/turbo/broadcastable.rb
```diff
@@ -51,15 +51,15 @@ module ClassMethods
     #     belongs_to :board
     #     broadcasts_to ->(message) { [ message.board, :messages ] }, inserts_by: :prepend, target: "board_messages"
     #   end
-    def broadcasts_to(stream, inserts_by: :append, target: broadcast_target_default)
-      after_create_commit  -> { broadcast_action_later_to stream.try(:call, self) || send(stream), action: inserts_by, target: target }
+    def broadcasts_to(stream, inserts_by: :append, target: broadcast_target_default, remove_if_present: nil)
+      after_create_commit  -> { broadcast_action_later_to stream.try(:call, self) || send(stream), action: inserts_by, target: target, remove_if_present: remove_if_present.nil? ? nil : remove_if_present.try(:call, self) || send(remove_if_present) }
```
jeromecornet, line 55: If we want this behaviour to be there always, it would be this:
```ruby
def broadcasts_to(stream, inserts_by: :append, target: broadcast_target_default)
      after_create_commit  -> { broadcast_action_later_to stream.try(:call, self) || send(stream), action: inserts_by, target: target, remove_if_present: self }
end
```
To make it opt-out instead of opt-in
```ruby
def broadcasts_to(stream, inserts_by: :append, target: broadcast_target_default, remove_if_present: :itself)
      after_create_commit  -> { broadcast_action_later_to stream.try(:call, self) || send(stream), action: inserts_by, target: target, remove_if_present: remove_if_present.blank? ? nil : remove_if_present.try(:call, self) || send(remove_if_present) }
```

with a similar change to `#broadcast`

tleish:
### Confusing/conflicting parameters
The proposed solution reads a bit confusing to 

```html
<ol id="list">
  <li id="item_1">Item 1</li>
  <li id="item_2">Item 2</li>
</ol>

<turbo-stream action="append" target="list" remove_if_present="item_2">
  <template>
    <li id="item_2">Item 2</li>
  </template>
</turbo-stream>
```
From a readability standpoint, it appears the code would 1) append the template content and then 2) remove item_2 if present.  Developer reading the code: "wait, why would I remove item_2 just after I append it?"

Is there a disadvantage to using existing functionality to accomplish the same thing?

## Existing Functionality
```html
<ol id="list">
  <li id="item_1">Item 1</li>
  <li id="item_2">Item 2</li>
</ol>

<turbo-stream action="remove" target="item_2"></turbo-stream>
<turbo-stream action="append" target="list">
  <template>
    <li id="item_2">Item 2 (UPDATED)</li>
  </template>
</turbo-stream>
```
This works if the item does or does not exist (no errors).  It also reads better.  1) remove existing item_2 and 2) append item_2.  

## Multiple Sub-IDs
The additional advantage of this approach is it supports more than one id.

```html
<ol id="list">
  <li id="item_1">Item 1</li>
  <li id="item_2">Item 2</li>
  <li id="item_3">Item 3</li>
</ol>

<turbo-stream action="remove" target="item_2"></turbo-stream>
<turbo-stream action="remove" target="item_3"></turbo-stream>
<turbo-stream action="append" target="list">
  <template>
    <li id="item_2">Item 2 (UPDATED)</li>
    <li id="item_3">Item 3 (UPDATED)</li>
  </template>
</turbo-stream>
```

## Consolidated Solution
If you wanted a more consolidate solution, you could do something like discussed in https://github.com/hotwired/turbo/issues/132.  

```diff
export const StreamActions: { [action: string]: (this: StreamElement) => void } = {
  append() {
+   this.removeDuplicateTargetChildrenById();
    this.targetElement?.append(this.templateContent)
  },

  prepend() {
+   this.removeDuplicateTargetChildrenById();
    this.targetElement?.prepend(this.templateContent)
  },

+ removeDuplicateTargetChildrenById() {
+   const templateChildrenIds = [...this.templateContent.children].map(child => child.id).filter(Boolean);
+   [...this.targetElement?.children].forEach(targetChild => {
+       if(templateChildrenIds.includes(targetChild.id)){
+         targetChild.remove();
+       }
+   });
+ }
```

## Jumping around
For me, the downside to any of these approaches is content jumping around.  Say you have an initial list:

- Item 1

Then 2 elements are appended to a list

- Item 1
- Item 2
- Item 3

Then item 2 is updated using action cable, but Item 3 is not.  

```html
<ol id="list">
  <li id="item_1">Item 1</li>
  <li id="item_2">Item 2</li>
  <li id="item_3">Item 3</li>
</ol>

<turbo-stream action="remove" target="item_2"></turbo-stream>
<turbo-stream action="append" target="list">
  <template>
    <li id="item_2">Item 2 (UPDATED)</li>
  </template>
</turbo-stream>
```

The the user perceives item 2 as randomly jumping to the bottom of the list.

- Item 1
- Item 3
- Item 2 (UPDATED)

## Replace/Append/Prepend
It might be better to replace existing elements when appending/prepending if they are present.  This would avoid the jumping around.  The code for that would be much more complicated.

jeromecornet:
Agreed on the jumping around.
My original version was replacing the element, but after @seanpdoyle 's comments re turbo.js I switched to remove/append rather than replace/append.

That said, I also agree that the code is confusing and the solution you highlighted dealing with this directly in turbo is much better, as it deals with it at the right layer.

I think I will close this and wait for hotwired/turbo#132.

Thanks for the detailed analysis!

tleish:
@jeromecornet — As far as I am aware, your solution is the only code towards resolving hotwired/turbo#132.  I'd be interested in you continuing your efforts, but I'm not on the core team.  @seanpdoyle - what are your thoughts on replace vs remove/append/prepend?

Issue #111, opened 2021-02-04
Installation fails in Rails Engine


vangberg:
Hi!

I am trying to install `turbo-rails` in a Rails Engine.

Steps to reproduce:

1. Create engine: `rails plugin new rails-turbo-engine --mountable`
2. Add `turbo-rails` to Gemfile (or gemspec, same issue)
3. Run `bin/rails turbo:install`

This fails with:

```
rails aborted!
Don't know how to build task 'turbo:install' (See the list of available tasks with `rails --tasks`)
Did you mean?  app:turbo:install
```

It seems like all tasks (including those provided by gems) are namespaced under `app:` in engines. So when I try `app:turbo:install` I get:

```
rails aborted!
Don't know how to build task 'turbo:install:asset_pipeline' (See the list of available tasks with `rails --tasks`)
Did you mean?  app:turbo:install:asset_pipeline

Tasks: TOP => app:turbo:install
(See full trace by running task with --trace)
```

Etc.

Repository for reproduction: https://github.com/vangberg/repro-rails-turbo-engine

FWIW, the same issue is present with tailwindcss-rails.

fuentesjr:
Same with hotwire-rails: https://github.com/hotwired/hotwire-rails/issues/22

dhh:
I didn't consider installing in an engine, so feel free to create a PR that rectifies that.

fuentesjr:
@dhh Ok I'll start with a PR for this, and if the approach looks good, I can do something similar for https://github.com/hotwired/hotwire-rails/issues/22

alec-c4:
hey, any news with this issue?

PR #162, opened 2021-04-17
Add support for installing via rails mounted engines

fuentesjr:
### Overview
This addresses issue #111 and will potentially help in addressing https://github.com/hotwired/hotwire-rails/issues/22 as well. This adds new tests that ensure the rake tasks are mounted/loaded in mounted engine applications and "typical" applications as well.


app/helpers/turbo/streams_helper.rb
```diff
@@ -40,8 +40,8 @@ def turbo_stream
   # (when Current.account.id = 5). Updates to this stream can be sent like
   # <tt>entry.broadcast_append_to entry.account, :entries, target: "entries"</tt>.
   def turbo_stream_from(*streamables, **attributes)
-    attributes["channel"] = "Turbo::StreamsChannel"
-    attributes["signed-stream-name"] = Turbo::StreamsChannel.signed_stream_name(streamables)
+    attributes[:channel] = "Turbo::StreamsChannel"
+    attributes[:"signed-stream-name"] = Turbo::StreamsChannel.signed_stream_name(streamables)
```
fuentesjr, line 44: I was seeing "TypeError: hash key "channel" is not a Symbol" failing/error tests on:

BroadcastsTest#test_Users::Profile_broadcasts_Turbo_Streams
BroadcastsTest#test_Message_broadcasts_Turbo_Streams
Turbo::StreamsHelperTest#test_with_streamable
Turbo::StreamsHelperTest#test_with_streamable_and_html_attributes

https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-tag suggests that these should be symbols, so after this change the tests were passing for me. I'm not sure why this wasn't failing before. If I missed something let me know and I can revert this change.

Gemfile.lock
```diff
@@ -175,6 +183,7 @@ DEPENDENCIES
   sqlite3
   turbo-rails!
   webdrivers
+  webpacker
```
fuentesjr, line 186: I added this dev dependency to test the `app:turbo:webpacker:install` task

lib/tasks/turbo_tasks.rake
```diff
@@ -1,24 +1,29 @@
-def run_turbo_install_template(path) system "#{RbConfig.ruby} ./bin/rails app:template LOCATION=#{File.expand_path("../install/#{path}.rb",  __dir__)}" end
+def run_turbo_install_template(path, nspace = nil)
+  system "#{RbConfig.ruby} #{Rails.root}/bin/rails #{nspace}app:template LOCATION=#{File.expand_path("../install/#{path}.rb",  __dir__)}"
+end
 
 namespace :turbo do
   desc "Install Turbo into the app"
-  task :install do
+  task :install do |task|
+    nspace = task.name.split(/turbo:install/).first
```
fuentesjr, line 8: This is ultimately the approach that I took to determine whether the rake task invocations should include the app namespace. An alternative approach I was considering is something like:

```ruby
nspace = nil
nspace = "app:" if defined?(ENGINE_ROOT) 
```
I came across `ENGINE_ROOT` from one of the [rails source files](https://github.com/rails/rails/blob/main/railties/lib/rails/tasks/engine.rake), but I don't know enough of rails to be confident that will always work.

fuentesjr (commented):
I added some small notes about my first approach here

danlaffan:
Hi @dhh and @fuentesjr ,

Any thoughts on this PR?  Any update would be great as I am starting a large project using Rails engines and I need to know if Hotwire will run properly from inside an engine.

Thanks everyone.

dhh:
@danlaffan Turbo works fine in an engine. This PR is just about setting up an install script. Everything this gem provides is automatically available everywhere.

danlaffan:
@dhh That's correct, David.  The installer script for turbo-rails doesn't work currently when installing in an engine.  However, I have proven that if you perform the steps manually, you definitely CAN get turbo to work in an engine.

FYI, the same problem (install script not working inside an engine) also pertains to the Stimulus-rails gem.  I intend to fix both of these in forthcoming pull requests, hopefully this week.

I noticed that the turbo gem (as I last saw it) doesn't create an importmap.js.erb file.  I intend to remedy that as well in my forthcoming pull request.

Cheers, 
Dan

romaingibs13:
Hello what is the current state of Hotwire on Rails Engine ? Because I am seeing issues as of now, here a screenshot:
Screenshot 2022-01-25 at 10 52 45

danlaffan:
Hi @romaingibs13,

I have modified 4 gems locally so they work in the context of a Rails 7 engine.  I will create pull requests for them today.  I have modified all of them in the same way, so if one gets approved, the others should too.  

They are:

* turbo-rails
* stimulus-rails
* hotwire-rails
* importmap-rails

I will start with a PR for importmap-rails.  I

romaingibs13:
@danlaffan Thanks a lot for your quickness of answer, happy to know it's already on the way :) We are looking forward to use this promising technology

danlaffan:
Check this out - https://github.com/rails/importmap-rails/pull/103 - all thoughts welcome.

WriterZephos:
Not sure if this will solve the issue I am facing, but when creating a new plugin via `rails plugin new` the dummy app that gets created does not have hotwire installed and I get the following error when trying to call `turbo_frame_request?`

```
Error:
ControllerTest#test_turbo_router_handles_standard_requests_by_rendering_default_template:
NoMethodError: undefined method `turbo_frame_request?' for #<DefaultTestController:0x00000000002f08>
```

Please let me know if this is the same issue. I am new to the whole plugin/engine thing.

WriterZephos:
@danlaffan can you outline the steps to get turbo-rails installed in a demo app for a plugin? This PR is taking a long time and would like to work around this problem for now.

danlaffan:
Hi @WriterZephos,

This pull request shows the basic steps I had to take to get the turbo-rails, stimulus-rails and hotwire-rails gems to work in the context of a Rails engine.  I have PRs ready to go for those other three gems when/if this PR is accepted.

Could you kindly clarify for me – do you intend to install hotwire-rails into:
* A plugin which is then included in a Rails app, OR
* Directly into the parent Rails app, but use it in the code inside the plugin?

Best regards, Dan.

WriterZephos:
> Hi @WriterZephos,
> 
> I have created a pull request in importmap-rails that shows the basic steps I had to take to get the turbo-rails, stimulus-rails and hotwire-rails gems to work in the context of a Rails engine. I have PRs ready to go for those other three gems when/if the importmap-rails PR is accepted.
> 
> Could you kindly clarify for me – do you intend to install hotwire-rails into:
> 
> * A plugin which is then included in a Rails app, OR
> * Directly into the parent Rails app, but use it in the code inside the plugin?
> 
> Best regards, Dan.

Hi @danlaffan 

I will take a look at that other MR. Thank you.

I want to include turbo-rails directly into the rails app. The gem I am making basically adds helpers and render methods to make using turbo-rails easier (they will use turbo-rails from inside the plugin code). I just need to be able to install it in the demo app so I can write tests.

In theory I can build my app without it but testing would be difficult.

Thanks

danlaffan:
Hi @WriterZephos,

If you wish to install turbo-rails direct into a Rails app, you should be able to directly add it to the gemfile of the Rails app.  My pull requests are to help people who are trying to install the turbo-rails gem directly in the plugin/Rails Engine.

However, if your gem/plugin requires turbo-rails, it would then become dependent on the turbo-rails gem, so perhaps you should be considering adding turbo-rails as a dependency into your plugin/gem.

Am I understanding you correctly?

Best regards, Dan

WriterZephos:
@danlaffan  I think I get what you are saying, but the problem isn't the dependency, it's the inability to test my gem because it requires being able to call turbo-rails methods from the controller.

One of the things my gem provides is a mixin/concern/module that you can include in your controllers, and that code calls turbo-rails methods. Without having turbo-rails installed in the demo app that is generated, I can't properly test my plugin code against it.

I can do manually testing, sure, because that would use a standalone app that wasn't automatically generated without turbo-rails.

danlaffan:
Hi @WriterZephos, 

Have you tried installing turbo-rails in your test app?  

In my current project (my first turbo project), I have a Rails Engine which has a folder called spec/dummy that contains a test Rails app.  When I run `bin/rails s` it runs the app in that spec/dummy folder.  You could add turbo into the gemfile for such an app, or you could make turbo a dependency of the plugin, and then your host app would need to / be required to load turbo.

How you actually trigger tests to call your turbo mini depends a lot on your code, and of course on which testing system you're using.  I'm really only familiar with RSpec.

Would you like to transfer this discussion to a Stack Overflow question where we can look at code samples?  If you would, paste in the link to your SO question here and I will pick this up tomorrow.  It's the end of the day here in Ireland, so I'll pick this up again tomorrow.

Best regards, Dan

WriterZephos:
Hey @danlaffan your comments got me thinking about it differently and I tried again. I was able to run `rails turbo:install` from the demo app root and it worked! Now I can test my gem with a turbo enabled demo app.

I really appreciate the conversation to help me unlock my brain.

One thing of note: I also had to install `jsbundling-rails` since turbo requires either node or importmaps to run, which was surprising but makes sense I suppose, since it comes with a JS library.

Thanks again. Cheers.

danlaffan:
No problem! :)

fuentesjr:
Hello @danlaffan 👋🏼 

I think this PR is a bit outdated, so I'm thinking it would be best to close this one. Thoughts?

danlaffan:
Hi @fuentesjr, I am happy to update it.  I think it's still needed.  Am I missing something?

fuentesjr:
Hey @danlaffan Ok let's update it and then I'll remove the Draft status. It was just starting to look like this PR was becoming stale.

fuentesjr:
@danlaffan Any update on this? Would you still like to keep this open?

danlaffan:
Hi @fuentesjr , I'm sorry for the confusion I have caused here.  Yes, go ahead and lose this PR.  I have created another PR in the importmap-rails gem as a starting point in fixing this issue.  This PR can be closed.

PR #194, opened 2021-06-30
Turbo::Streams::TagBuilder can set targets attribute on stream tag

t27duck:
Background: https://github.com/hotwired/turbo/pull/113 allowed for multiple targets in turbo frame streams. This attempts to expose the feature to the Rails turbo_stream tag helper. Instead of a `target` which results in a call to `document.getElementById` on the JS level, a `targets` attribute will be passed into `document.querySelectorAll`

This adds the functionality to the tag builder. All "action methods" now have an `_all` variant which is piped into a new `action_all` method. The only real difference in these methods is the end result is setting `targets` instead of `target` in `turbo_stream_action_tag` which has also been updated to handle this feature.

I purposefully did not touch the ActionCable-related part of the gem as I think its usecase is mostly geared to always updating one element at a time. Regardless, if it were to get this feature, I think it would be better in a separate pull request. For now, I think giving the developer the ability to target multiple elements at once in the turbo stream response from a controller is ok for now.

The test included is pretty much a copy/paste of an existing test, but invoking the new code path.

### DIFF
--- a/app/models/turbo/streams/tag_builder.rb
+++ b/app/models/turbo/streams/tag_builder.rb
@@ -40,6 +40,16 @@ class Turbo::Streams::TagBuilder
     action :remove, target, allow_inferred_rendering: false
   end
 
+  # Removes the <tt>targets</tt> from the dom. The targets can either be a CSS selector string or an object that responds to
+  # <tt>to_key</tt>, which is then called and passed through <tt>ActionView::RecordIdentifier.dom_id</tt> (all Active Records
+  # do). Examples:
+  #
+  #   <%= turbo_stream.remove_all ".clearance_item" %>
+  #   <%= turbo_stream.remove_all clearance %>
+  def remove_all(targets)
+    action_all :remove, targets, allow_inferred_rendering: false
+  end
+
   # Replace the <tt>target</tt> in the dom with the either the <tt>content</tt> passed in, a rendering result determined
   # by the <tt>rendering</tt> keyword arguments, the content in the block, or the rendering of the target as a record. Examples:
   #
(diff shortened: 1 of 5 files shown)


app/models/turbo/streams/tag_builder.rb
```diff
@@ -32,108 +32,127 @@ def initialize(view_context)
...
   # <tt>to_key</tt>, which is then called and passed through <tt>ActionView::RecordIdentifier.dom_id</tt> (all Active Records
-  # do). Examples:
+  # do), or a CSS selector string (when using <tt>target_multiple: true</tt>). Examples:
   #
   #   <%= turbo_stream.remove "clearance_5" %>
   #   <%= turbo_stream.remove clearance %>
-  def remove(target)
-    action :remove, target, allow_inferred_rendering: false
+  #   <%= turbo_stream.remove ".invalid-row", target_multiple: true %>
+  def remove(target, target_multiple: false)
```
dhh, line 40: I'd rather see us add a separate `#remove_all(targets)` method than overloading this one.

app/models/turbo/streams/tag_builder.rb
```diff
@@ -32,108 +32,127 @@ def initialize(view_context)
...
   #   <%= turbo_stream.before "clearance_5", "<div id='clearance_4'>Insert before the dom target identified by clearance_5</div>" %>
   #   <%= turbo_stream.before clearance %>
   #   <%= turbo_stream.before clearance, partial: "clearances/clearance", locals: { title: "Hello" } %>
   #   <%= turbo_stream.before "clearance_5" do %>
   #     <div id='clearance_4'>Insert before the dom target identified by clearance_5</div>
   #   <% end %>
-  def before(target, content = nil, **rendering, &block)
-    action :before, target, content, **rendering, &block
+  #   <%= turbo_stream.before ".item", target_multiple: true, partial: "clearances/clearance_notification", locals: { title: "Hello" } %>
+  def before(target, content = nil, target_multiple: false, **rendering, &block)
```
dhh, line 70: Same as with remove. Add a `before_all`.

app/models/turbo/streams/tag_builder.rb
```diff
@@ -32,108 +32,127 @@ def initialize(view_context)
...
-    action :prepend, target, content, **rendering, &block
+  #   <%= turbo_stream.prepend ".item", target_multiple: true, partial: "clearances/clearance_notification", locals: { title: "Hello" } %>
+  def prepend(target, content = nil, target_multiple: false, **rendering, &block)
+    action :prepend, target, content, target_multiple: target_multiple, **rendering, &block
   end
 
   # Send an action of the type <tt>name</tt>. Options described in the concrete methods.
-  def action(name, target, content = nil, allow_inferred_rendering: true, **rendering, &block)
-    target_name = extract_target_name_from(target)
+  def action(name, target_or_targets, content = nil, target_multiple: false, allow_inferred_rendering: true, **rendering, &block)
```
dhh, line 137: `action_all` also.

dhh (changes requested):
Let's go with separate method calls rather than overloading existing ones.

t27duck:
> Let's go with separate method calls rather than overloading existing ones.

I like that idea. I'll see about getting that implemented here in the near future...

t27duck:
Reimplemented the PR where all the action methods now have an `_all` version. I also updated the commit message and PR description to reflect what's actually happening now. I'm liking this change much more than my original idea.

PR #203, opened 2021-07-16
Turbo frame identifier prefixing

ayrton:
## What

Add a little helper to easily prefix turbo frame identifiers.

## Why

Your models don't always live in a single top-level namespace. Eg in some applications a resource can be both private and publicly accessible:

```rb
# config/routes.rb

namespace :public do
  resources :posts, only: :show
end

namespace :private do
  resources :posts, only: :show
end
```

These resources can look visually differently or behave differently so it can make sense to target these by a different identifier.

## Example

__Before:__

```erb
<%= turbo_frame_tag dom_id(message, :shared) do %>
  <article></article>
<% end %>
```

__After:__

```erb
<%= turbo_frame_tag message, :shared do %>
  <article></article>
<% end %>
```


dhh:
I see the appeal, but don't love this being an unnamed second argument. I tried a few different ways to see if there was something clearer than simply calling dom_id, and I couldn't find that.

PR #205, opened 2021-07-21
Add broadcast_update functionality

### LINKED ISSUE #204: Why is there no `Turbo::StreamsChannel.broadcast_update_to`
The `update` action is unavailable when broadcasting stream actions via websockets. It's possible to workaround by using `replace` instead (and configuring partials accordingly), but is it possible instead to add `broadcast_update_to` to this [`Turbo::Streams::Broadcasts`](https://github.com/hotwired/turbo-rails/blob/main/app/channels/turbo/streams/broadcasts.rb#L12-L14)?

efatsi:
Resolves #204.

More or less just copy-pasting the existing `broadcast_replace` functionality with `broadcast_update` counterparts

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ Turbo can coexist with Rails UJS, but you need to take a series of upgrade steps
 ## Development
 
 * To run the Rails tests: `bundle exec rake`.
+  * To install dependencies: `bundle install`
+  * To prepare the test database: `cd tests/dummy; RAILS_ENV=test bundle exec rails db:migrate`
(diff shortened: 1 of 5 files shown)


dhh:
There's a test failure re: some naming.

efatsi:
👍 will push through test failures, sorry for the noise

efatsi:
Specs passing locally, added some notes to the readme to help future devs run those. LMK if I should split that out into a separate PR, or just drop it.

PR #217, opened 2021-08-19
Add broadcast remove later functionality

cpjmcquillan:
Add methods to broadcast remove actions asynchronously like you can with
the other actions.

### DIFF
--- a/app/channels/turbo/streams/broadcasts.rb
+++ b/app/channels/turbo/streams/broadcasts.rb
@@ -39,6 +39,9 @@ module Turbo::Streams::Broadcasts
     )
   end
 
+  def broadcast_remove_later_to(*streamables, target:)
+    broadcast_action_later_to *streamables, action: :remove, target: target
+  end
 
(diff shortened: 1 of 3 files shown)


dhh:
Remove doesn't require any rendering, so there's no reason to run it async. That's why it wasn't there in the first place.

Have you discovered a different need?

cpjmcquillan:
> Remove doesn't require any rendering, so there's no reason to run it async. That's why it wasn't there in the first place.
> 
> Have you discovered a different need?

In the scenario where a user interaction triggers several different types of broadcasts to a page, the remove sticks out from the rest of the async updates. 

Perhaps this is a design/UX problem more than anything else, though. 

Feel free to close if that's not a compelling enough reason to add it!

dhh:
Gotcha, yeah, I think it's fine for it to stick out. The only reason we do async is because there's rendering going on. So it's appropriate that remove is different from actions that need rendering 👍

Issue #225, opened 2021-08-26
Automatically run turbo:install:redis if Redis is available on the machine already


dhh:
You need Redis to use Turbo Streams. Would be nicer if it was turned on by default during turbo:install if Redis is detected on the system.

ceritium:
https://github.com/hotwired/turbo-rails/pull/226 address it

PR #226, opened 2021-08-26
Switch on Redis if available on turbo:install

ceritium:
It checks (naively) if Redis is installed, in case affirmative it run `turbo:install:redis`, otherwise print some information about how to do it.

Closes https://github.com/hotwired/turbo-rails/issues/225


mdavidn:
This prompted me to uninstall Redis locally. I work on multiple Rails projects, but only a handful have Redis available in production.

PR #230, opened 2021-09-10
Include `#` prefix in `[targets]`

seanpdoyle:
Since `<turbo-stream targets="...">` values are CSS selectors and _not_
element identifiers, mapping a collection of elements that integrate
with `dom_id` results in unresolvable selectors.

For example, passing an array of `Message` instances:

```ruby
messages = [ 1, 2, 3 ].map { |id| Message.new(id: id) }

turbo_stream.replace_all messages #=> <turbo-stream targets="message_1 message_2 message_3">...</turbo-stream>
```

When mapping a collection of targets that respond to `to_key` and
integrate with `dom_id`, map each value and prefix it with the [ID
selector][].

As a result, the `[targets]` attribute will be a space-separated list of
valid ID selectors:

```ruby
messages = [ 1, 2, 3 ].map { |id| Message.new(id: id) }

turbo_stream.replace_all messages #=> <turbo-stream targets="#message_1 #message_2 #message_3">...</turbo-stream>
```

[ID selector]: https://developer.mozilla.org/en-US/docs/Web/CSS/ID_selectors

### DIFF
--- a/app/helpers/turbo/streams/action_helper.rb
+++ b/app/helpers/turbo/streams/action_helper.rb
@@ -14,7 +14,7 @@ module Turbo::Streams::ActionHelper
 
     if target = convert_to_turbo_stream_dom_id(target)
       %(<turbo-stream action="#{action}" target="#{target}">#{template}</turbo-stream>).html_safe
-    elsif targets = convert_to_turbo_stream_dom_id(targets)
+    elsif targets = convert_to_turbo_stream_dom_id(targets, include_selector: true)
       %(<turbo-stream action="#{action}" targets="#{targets}">#{template}</turbo-stream>).html_safe
     else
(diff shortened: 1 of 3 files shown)


4lllex:
Appears this doesn't work anymore. 
```rb
>> messages = [1, 2].map { |id| Message.new(id: id) }
>> helper.turbo_stream.replace_all messages
=> "<turbo-stream action=\"replace\" targets=\"##&lt;Message:0x00007fb0031c8ad0&gt;_message_1\"><template></template></turbo-stream>"
```

It also doesn't work as a selector in turbo without a comma: `document.querySelectorAll("#message_1 #message_2")` 
*https://github.com/hotwired/turbo/blob/fdd11dc968a86bfc1839d2d4387118a114117436/src/elements/stream_element.js#L182*

Issue #229, opened 2021-09-08
Turbo-frame request variant


tleish:
Any interest in adding a turbo-frame request variant to `Turbo::Frames::FrameRequest`?

```diff
# turbo-rails/app/controllers/turbo/frames/frame_request.rb

module Turbo::Frames::FrameRequest
  extend ActiveSupport::Concern

  included do
    layout -> { false if turbo_frame_request? }
    etag { :frame if turbo_frame_request? }
+   before_action :turbo_frame_request_variant
  end

  private
    def turbo_frame_request?
      request.headers["Turbo-Frame"].present?
    end

+   def turbo_frame_request_variant
+     request.variant = :turbo_frame if turbo_frame_request?
+   end
end
```

This provides the option to change behavior on controller actions for turbo-frame requests:

```ruby
class PostsController < ApplicationController
  def new
    @post = Post.new
    respond_to do |format|
      format.html do |variant|
        variant.turbo_frame # renders app/views/posts/new.html+turbo_frame.erb
        variant.none # renders app/views/posts/new.html.erb
      end
    end
  end

  def create
    respond_to do |format|
      format.turbo_stream do |variant|
        variant.turbo_frame { redirect_to ... } # turbo-frame specific behavior
        variant.none { redirect_to ... }
      end
      format.html { redirect_to ... }
    end
  end
end
```

see also: https://github.com/hotwired/turbo/issues/378

WriterZephos:
I strongly recommend doing this. It doesn't make a lot of sense to not have an easy way to respond differently to from requests coming externally, and requests coming from internal links. For example, if someone types in a url to the show view of some model, you want to render with a template, but if you are responding in the controller to render a partial for internal navigation, they would only get that.

Since I don't like having a bunch of different views that only differ in their file extensions, I like to render the same view and use different templates for replacing turbo frames in the dom or rendering the full page. This works great!

WriterZephos:
Just wanted to drop a link to the gem I am working on that would benefit from this change. The module here dynamically selects a layout based on the request variant:

https://github.com/WriterZephos/TurboRouter/blob/main/lib/turbo_router/controller.rbz

I am rebuilding the above repo using `rails plugin new` instead of `bundle gem` but the demo app that gets generated for testing doesn't have hotwire installed, so I am waiting for that to be fixed.

SleeplessByte:
[...]

PR #232, opened 2021-09-10
Include layout for `Turbo-Frame:` requests

seanpdoyle:
Include layout for `Turbo-Frame:` requests
===

The bandwidth benefits of optimizing `Turbo-Frame:` header responses to
omit the layout do not offset the HTTP-layer cache busting trade-offs.

Similarly, there's an opportunity to implement behavior changes related
to:

* [@hotwired/turbo#361][]
* [@hotwired/turbo#257][]

if both `<turbo-frame>` elements' `FrameController` instances and
`Session` instances were able to share `Visit` instances built from
fully formed `<html>` pages. By trimming the layout and outer portions
of the document, `turbo-rails` forces frames to deal with incomplete
fragments.

[@hotwired/turbo#257]: https://github.com/hotwired/turbo/issues/257
[@hotwired/turbo#361]: https://github.com/hotwired/turbo/issues/361

Mark variant as `turbo_frame` when header present
===

Closes https://github.com/hotwired/turbo-rails/issues/229

---

Mark the request's [variant][] as `:turbo_frame` whenever the
`Turbo-Frame:` header is present in the request. Also adds documentation
to describe how to achieve the previously built-in layout optimizations.

[variant]: https://guides.rubyonrails.org/4_1_release_notes.html#action-pack-variants


dhh:
I like the idea of simplifying things, but there are more considerations than just bandwidth. There's also the price of producing the request server side. Like the cost to produce the layout itself. But... that cost is paid everywhere else, so is that enough?

I'd like to verify this on our apps though.

Also, there's the slight regression in readability on the return when you're debugging. It's been very nice just to see the frame that changed.

seanpdoyle:
If https://github.com/hotwired/turbo/pull/398 is viable, it'd be simpler to traffic in full HTML documents.

If there's interest to support `<turbo-frame>`-only responses without the outer document, https://github.com/hotwired/turbo/pull/398 would need to synthesize the `VisitOptions.responseHTML` by injecting the `<turbo-frame>` into the browser's "live" document, then transforming the DOM into an HTML string.

It would be possible with only fragments, but committing to transmitting full HTML documents as responses has other upsides that make it seem worthwhile.

This is, of course, only a requirement for frames that drive the browser's history. If applications or frames aren't interested in that, they can opt-into the bare `<%= yield %>` layout via the `html+turbo_frame` variant.

It would be easier to opt-in for scenarios that need it than it is to reverse the existing default behavior of never sending a full document response.

dhh:
So been thinking about this some more. I don't think layout-by-default is going to give us any caching advantages, because most turbo frames are used/loaded exclusively for that purpose. They're rarely used directly. At least that's the usage across all of HEY for us.

Furthermore, many layouts are actually pretty big. For HEY, now that we're going with import maps, there's a lot of stuff in the head. Don't want to keep shipping that across all the time.

app/controllers/turbo/frames/frame_request.rb
```diff
@@ -1,20 +1,24 @@
...
-# <tt>Turbo-Frame</tt> header to the request. When that header is detected by the controller, we ensure that any
-# template layout is skipped (since we're only working on an in-page frame, thus can skip the weight of the layout), and
-# that the etag for the page is changed (such that a cache for a layout-less request isn't served on a normal request
-# and vice versa).
+# <tt>Turbo-Frame</tt> header to the request.
 #
-# This is merely a rendering optimization. Everything would still work just fine if we rendered everything including the layout.
-# Turbo Frames knows how to fish out the relevant frame regardless.
+# When the <tt>Turbo-Frame</tt> header is present, the request's <a
+# href="https://guides.rubyonrails.org/4_1_release_notes.html#action-pack-variants">variant</a>
```
ghiculescu, line 6:
```suggestion
# href="https://guides.rubyonrails.org/layouts_and_rendering.html#the-variants-option">variant</a>
```

ghiculescu:
Something that maybe hasn't been noted yet: At the moment if you set `layout "whatever"` in a controller that's also receiving frame requests, the current optimisations are lost (since you've overridden the `layout`). It looks like this PR presents a way to fix that, which is awesome.

app/controllers/turbo/frames/frame_request.rb
```diff
@@ -1,20 +1,24 @@
...
+# will be marked <tt>turbo_frame</tt>.
+#
+# As an optimization, applications can declare a layout with the
+# <tt>html+turbo_frame.erb</tt> extension that renders
+# without the navigation and layout elements:
+#
+#     <%= app/views/layouts/application.html+turbo_frame.erb %>
```
ghiculescu, line 13:
```suggestion
#     <%# app/views/layouts/application.html+turbo_frame.erb %>
```

songjiz:
> Something that maybe hasn't been noted yet: At the moment if you set `layout "whatever"` in a controller that's also receiving frame requests, the current optimisations are lost (since you've overridden the `layout`). It looks like this PR presents a way to fix that, which is awesome.

```ruby
# app/controllers/application_controller.rb
before_action -> { self.class.layout(false) if turbo_frame_request? }
```

It's works for me.

ghiculescu:
Ahhh, sorry, I meant if you do something like this

```ruby
class ApplicationController
  layout "my_layout_view"
end
```

Then that overrides the layout that `turbo-rails` adds, so your frame requests get a layout.

This PR fixes it though so that's cool. Also you can just take the stuff in this PR and copy it into your own app (which I did yesterday and it worked great).

app/controllers/turbo/frames/frame_request.rb
ghiculescu: Would it be possible to generate this view as part of the install generator?

seanpdoyle:
For what it's worth, the `turbo_frame` variant support is also covered by https://github.com/hotwired/turbo-rails/pull/250.

The behavior mentioned in https://github.com/hotwired/turbo-rails/pull/232#issuecomment-929587457 relies on the variant portion of these changes more than reverting the `before_action` and `etag` blocks.

WriterZephos:
I don't thins this Closes  exactly. The solution proposed in https://github.com/hotwired/turbo-rails/issues/229 is much more flexible in my opinion, and I think not sending the full rendered page on every request is a major benefit and one of turbo's strengths. It would be a shame to lose that.

I favor the solution in provided by https://github.com/hotwired/turbo-rails/pull/250

seanpdoyle:
I've opened https://github.com/hotwired/turbo/pull/445 to introduce the `turbo:frame-missing` event to provide applications with a seam to respond to a request when the response does not contain a matching `<turbo-frame>` element.

In some cases, (like in the case of a `401...403` status code in response to an expired session), applications might want to replace the contents of the current page with the frame-less Response HTML.

That capability is contingent on the response being a fully formed HTML document.

There's also an ongoing discussion about providing applications with the ability to ["break out" of a frame](https://github.com/hotwired/turbo/issues/257#issuecomment-968179718). The proposed version of that feature would also depend on the response to a request with the`Turbo-Frame:` header containing a fully-formed HTML document.

WriterZephos:
> Something that maybe hasn't been noted yet: At the moment if you set `layout "whatever"` in a controller that's also receiving frame requests, the current optimisations are lost (since you've overridden the `layout`). It looks like this PR presents a way to fix that, which is awesome.

So I think setting a layout globally SHOULD override the current behavior. I think if you want to set a layout in a more nuanced way, you need a before action such as what was proposed above or something like the pattern I implemented here:

https://github.com/WriterZephos/TurboRouter/blob/main/lib/turbo_router/controller.rb

app/controllers/turbo/frames/frame_request.rb
WriterZephos: What exactly would be in this layout? Seems like we would need a way to customize the turbo_frame_id here if it wraps the rendered content with a turbo_frame, but I am not sure that is the intent.

bborn:
Chiming in here in favor of rendering with the layout (or at least making this configurable). If you have a layout that uses a `content_for` block, this causes lots of confusion. For example in my case, my layout has something like 

```ruby
-if content_for(:sidebar)
  = yield :sidebar
```

If I try to render a turbo response with a template that captures the `:sidebar` content, it just doesn't work, because the layout isn't rendered at all.

dhh:
I don't think doing the layout for all turbo requests is the right trade-off as a default. But I'd be happy to take a way, or perhaps just documentation, to let people tweak this for an individual app. Let's explore that in a new PR, though.

seanpdoyle:
A new edge case has been uncovered that hasn't been mentioned in this discussion: [when responding to a form submitted from within a `<turbo-frame>` with a Turbo Stream response, the `app/views/layouts/application.turbo_stream.erb` layout is omitted](https://github.com/hotwired/turbo-rails/pull/351). Circumventing the default `layout -> { false if turbo_frame_request? }` block becomes tricky, since the response's MIME type isn't known until [very late in the controller's action](https://github.com/hotwired/turbo-rails/pull/351#discussion_r913851213).

For example, a controller action "ending" with `redirect_to(...)` would respond with an HTML content type redirect response without a layout, but one that "ends" with

```ruby
respond_to do |format|
  format.html { redirect_to(...) }
  format.turbo_stream
end
```

would respond with a Turbo Stream content type, and that response would **surprisingly** omit a layout.

If the `layout { false if turbo_stream_request? }` block were removed, then responses to Turbo Frame HTML requests would include a layout **and** responses to Turbo Stream requests would also include a layout.

dhh:
Wouldn't this fly:

```ruby
respond_to do |format|
  format.html { redirect_to(...) }
  format.turbo_stream { render layout: "something" }
end
```

?

Kulgar:
Same problem as @seanpdoyle 

@dhh yes it would (and it works), but then, we would have to put the "render layout" every time? Or is there a better way?

PR #235, opened 2021-09-16
Added turbo_frame_id method to return "Turbo-Frame" header value

wollistik:
Shortcut when in need for the id of the requesting turbo frame.
We use it in our default error handling to generate an appropriate turbo_frame tag, which will display the error.


dhh:
Followed up with 48d40d49c26e4be252110eee93597a4a31dbf42a.

PR #238, opened 2021-09-17
Make `bin/rails` executable

seanpdoyle:
Make `bin/rails` executable:
===

This commit contains the results of:

```sh
chmod a+x bin/rails
```

Use `bin/test` in CI
===

Use `bin/test` with a pattern for finding files instead of
`bundle exec rake`. We're passing a pattern to `bin/test` because
without it, tests in `test/system` are skipped:

For example:

```
~/Work/hotwired/turbo-rails bin-rails-execution-permissions*                                                              [7/2071]
❯ bin/test 
Installing Ruby dependencies
Warning: the running version of Bundler (2.2.12) is older than the version that created the lockfile (2.2.16). We suggest you to u
pgrade to the version that created the lockfile by running `gem install bundler:2.2.16`.
Installing JavasScript

app/javascript/turbo/index.js → app/assets/javascripts/turbo.js...
created app/assets/javascripts/turbo.js in 692ms
Migrating test database
Run options: --seed 2063

# Running:

.............................................................

Finished in 0.612377s, 99.6118 runs/s, 174.7290 assertions/s.
61 runs, 107 assertions, 0 failures, 0 errors, 0 skips

~/Work/hotwired/turbo-rails bin-rails-execution-permissions* 7s
❯ gs
## bin-rails-execution-permissions
 M test/system/broadcasts_test.rb

~/Work/hotwired/turbo-rails bin-rails-execution-permissions*
❯ gd
diff --git a/test/system/broadcasts_test.rb b/test/system/broadcasts_test.rb
index 38a349e..9aa6e54 100644
--- a/test/system/broadcasts_test.rb
+++ b/test/system/broadcasts_test.rb
@@ -2,6 +2,7 @@ require "application_system_test_case"
  
 class BroadcastsTest < ApplicationSystemTestCase
   test "Message broadcasts Turbo Streams" do
+    fail
     visit messages_path
```

Since `bin/test` has the potential to generate changes to the embedded
assets, we'll fail CI if there are new and uncommitted changes after the
suite completes.


dhh:
Shouldn’t we use bin/test here?

PR #239, opened 2021-09-17
Support `[formmethod]` overrides to `_method`

seanpdoyle:
This commit translates the Turbo-internal changes proposed by
[hotwired/turbo#158][] to Turbo Rails via an event listener.

To quote the original pull request:

> Closes https://github.com/hotwired/turbo/issues/154.
> Closes https://github.com/hotwired/turbo/pull/155.
>
> To match the Rails idiom of override a form's `method` to support
> requests that are not `GET` and `POST`, this commit adds test coverage
> for existing functional behavior, along with some edge cases:
>
> * considering a submitter's `formmethod` value over a `<input
>   type="hidden" name="_method" value="...">` value
>
> * considering a submitter's `formmethod` value over a competing
>   submitter `name="_method" value="..."` pairing

[hotwired/turbo#158]: https://github.com/hotwired/turbo/pull/158


dhh:
Sean, what's the reason you wanted to move this over to turbo-rails rather than keep it in turbo? What's Rails specific about this?

seanpdoyle:
Relying on the combination of a `POST` verb and a `_method` body or query parameter is a Rails idiom and work-around, and isn't universal or part of the HTTP specification: 

https://edgeguides.rubyonrails.org/form_helpers.html#how-do-forms-with-patch-put-or-delete-methods-work-questionmark

seanpdoyle:
Declaring `<button name="_method" value="delete">Delete</button>` should currently work, outside the context of Turbo or Turbo-Rails.

There were changes introduced in https://github.com/rails/rails/pull/41045 that turn `formmethod:` options passed to `form.button` into `[name="_method"]` and `[value="delete"]` pairings so that those verbs work even in the absence of JavaScript.

PR #240, opened 2021-09-17
Run bundle after uncomment redis gem

ceritium:
Fix https://github.com/rails/rails/issues/43243

### DIFF
--- a/lib/install/turbo_needs_redis.rb
+++ b/lib/install/turbo_needs_redis.rb
@@ -1,6 +1,7 @@
 if (cable_config_path = Rails.root.join("config/cable.yml")).exist?
   say "Enable redis in bundle"
   uncomment_lines "Gemfile", /gem ['"]redis['"]/
+  run_bundle
 
   say "Switch development cable to use redis"
   gsub_file cable_config_path.to_s, /development:\n\s+adapter: async/, "development:\n  adapter: redis\n  url: redis://localhost:6379/1"


ghiculescu:
Could/should this also handle the case where redis isn’t in the gemfile?

ceritium:
In theory, "redis" is added to the Gemfile as part of cable ready.

https://github.com/rails/rails/blob/ecbd92cf48c741f784c1297041d368c8eb14e9bf/railties/lib/rails/generators/app_base.rb#L341in 

But it looks a bit weak and could fail if someone or something removes the gem. If @dhh agrees, then I can handle it in a different PR.

dhh:
Should handle the case where redis isn't listed in the Gemfile as well. You might well be adding turbo-rails to an existing app that maybe got started off --minimal 👍

ceritium:
Ok, I will take care of it.

gtoroap:
I'm still getting the same issue.

Rails 7.0.4
Bundler 2.3.7
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-darwin21]

```
Enable redis in bundle
        gsub  Gemfile
         run  bundle install
[40160, #<Thread:0x000000010a688cf0 run>, #<NameError: uninitialized constant Gem::Source

      (defined?(@source) && @source) || Gem::Source::Installed.new
                                           ^^^^^^^^
Did you mean?  Gem::SourceList> 
.
.
.

Switch development cable to use redis
        gsub  config/cable.yml
Could not find gem 'redis (~> 4.0)' in locally installed gems.
Run `bundle install` to install missing gems.

```

rreckonerr:
Same here!
rails 7.0.7.2
bundler 2.3.7
ruby 3.1.2p20

PR #245, opened 2021-09-23
Broadcast HTML from Model

yunggindigo:
Allows option to pass in html to broadcast from model
something like this will allow you to add html to a broadcast instead of a partial or otherwise automatically finding a partial with to_partial_path
`broadcast_replace_to(:feed, html: "<p> Random HTML </p>")`

Allows option to pass in html to broadcast from model this works with the **rendering methods included from the Turbo::Broadcastable module.

### DIFF
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -27,7 +27,7 @@
 # (which is derived by default from the plural model name of the model, but can be overwritten).
 #
 # You can also choose to render html instead of a partial inside of a broadcast
-# you do this by passing an html: option to any broadcast method that accepts **rendering options
+# you do this by passing the html: option to any broadcast method that accepts the **rendering argument
 # 
 #   class Message < ApplicationRecord
 #     belongs_to :user
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -296,9 +296,13 @@ module Turbo::Broadcastable
       options.tap do |o|
         # Add the current instance into the locals with the element name (which is the un-namespaced name)
         # as the key. This parallels how the ActionView::ObjectRenderer would create a local variable.
-        return if o[:html].present? # Allows html to be passed in instead of partial
-        o[:locals] = (o[:locals] || {}).reverse_merge!(model_name.element.to_sym => self)
-        o[:partial] ||= to_partial_path
+        if o[:html].present? 
+          # Allows html to be passed in instead of partial
+          o[:html] = o[:html]
+        else
+          o[:locals] = (o[:locals] || {}).reverse_merge!(model_name.element.to_sym => self)
+          o[:partial] ||= to_partial_path
+        end
       end
     end
 end
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -310,7 +310,7 @@ module Turbo::Broadcastable
       options.tap do |o|
         # Add the current instance into the locals with the element name (which is the un-namespaced name)
         # as the key. This parallels how the ActionView::ObjectRenderer would create a local variable.
-        # Dont Override HTML and Inline options by :partial
+        # if the html or inline options are passed it will skip setting a partial from #to_partial_path
         unless o.include?(:html) || o.include?(:inline)
           o[:locals] = (o[:locals] || {}).reverse_merge!(model_name.element.to_sym => self)
           o[:partial] ||= to_partial_path
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -310,9 +310,9 @@ module Turbo::Broadcastable
       options.tap do |o|
         # Add the current instance into the locals with the element name (which is the un-namespaced name)
         # as the key. This parallels how the ActionView::ObjectRenderer would create a local variable.
+        o[:locals] = (o[:locals] || {}).reverse_merge!(model_name.element.to_sym => self)
         # if the html or inline options are passed it will skip setting a partial from #to_partial_path
         unless o.include?(:html) || o.include?(:inline)
-          o[:locals] = (o[:locals] || {}).reverse_merge!(model_name.element.to_sym => self)
           o[:partial] ||= to_partial_path
         end
       end
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -26,6 +26,20 @@
 # and finally prepend the result of that partial rendering to the target identified with the dom id "clearances"
 # (which is derived by default from the plural model name of the model, but can be overwritten).
 #
+# You can also choose to render html or inline erb instead of a partial inside of a broadcast
+# you do this by passing an html: option or inline: to any broadcast method that accepts **rendering options
+# 
+#   class Message < ApplicationRecord
+#     belongs_to :user
+#
+#     after_create_commit :update_message_count
+#
+#     private
+#       def update_message_count
+#         broadcast_update_to(user, :message_count, html: "<p> #{user.messages.count} </p>")
+#       end
+#   end
+# 
 # There are four basic actions you can broadcast: <tt>remove</tt>, <tt>replace</tt>, <tt>append</tt>, and
 # <tt>prepend</tt>. As a rule, you should use the <tt>_later</tt> versions of everything except for remove when broadcasting
 # within a real-time path, like a controller or model, since all those updates require a rendering step, which can slow down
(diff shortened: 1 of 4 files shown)


dhh:
Nice! Can you amend docs too?

app/models/concerns/turbo/broadcastable.rb
```diff
@@ -296,8 +296,13 @@ def broadcast_rendering_with_defaults(options)
       options.tap do |o|
         # Add the current instance into the locals with the element name (which is the un-namespaced name)
         # as the key. This parallels how the ActionView::ObjectRenderer would create a local variable.
-        o[:locals] = (o[:locals] || {}).reverse_merge!(model_name.element.to_sym => self)
-        o[:partial] ||= to_partial_path
+        if o[:html].present?
```
seanpdoyle, line 299: I'd imagine that passing `html:` as an option is an edge case already, but it's worth highlighting that `html:` expects [valid and already escaped HTML text](https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-html):

> You can send an HTML string back to the browser by using the :html option to render:
>
>     render html: helpers.tag.strong('Not Found')
>
> This is useful when you're rendering a small snippet of HTML code. However, you might want to consider moving it to a template file if the markup is complex.
> 
> When using html: option, HTML entities will be escaped if the string is not composed with html_safe-aware APIs.

If you want to use templating or mix helper calls with hard coded HTML `String` instances, the [`inline:` option](https://guides.rubyonrails.org/layouts_and_rendering.html#using-render-with-inline supports ERB:

> The render method can do without a view completely, if you're willing to use the :inline option to supply ERB as part of the method call. This is perfectly valid:
>
>    render inline: "<% products.each do |p| %><p><%= p.name %></p><% end %>"
>
> There is seldom any good reason to use this option. Mixing ERB into your controllers defeats the MVC orientation of Rails and will make it harder for other developers to follow the logic of your project. Use a separate erb view instead.

With that being said, would this commit's tests pass _without_ this implementation change? `o[:html] = o[:html]` is a no-operation, and the code without these changes already passes the `o` Hash through to the underlying `render` call.

Am I misunderstanding something else at play here?

yunggindigo: Yes this is because by default if you didn't pass in a partial it will use #to_partial_path on the object with these changes you would be able to pass in html: "<p> </p>" and this would get rendered currently it will throw an error couldn't find partial. This is going to be useful for use with view_component as well because you can just render the view_component and stick it as the html argument

seanpdoyle: I understand now, the crucial part of the proposed change is the conditional, and not the contents of the block.

In that case, what do you think about guarding the `locals` and `partial` assignment with something more explicit, and omitting the `o[:html] = o[:html]`? Maybe something like:

```ruby
unless o.include?(:html) || o.include?(:inline)
  o[:locals] = (o[:locals] || {}).reverse_merge!(model_name.element.to_sym => self)
  o[:partial] ||= to_partial_path
end
```

yunggindigo: Right I see what you mean that would work well as it just passes the hash along

yunggindigo: also with the added bonus of supporting inline right away

test/system/broadcasts_test.rb
```diff
@@ -10,6 +10,19 @@ class BroadcastsTest < ApplicationSystemTestCase
...
+    visit messages_path
+
+    assert_text "Messages"
+    message = Message.new(record_id: 1, content: "A new message")
+    message.broadcast_append_to(:messages)
+    assert_text("A new message")
+    
+    message.broadcast_update_to(:messages, target: "messages-count", 
+      html: "#{Message.count} messages sent")
+    assert_selector("#messages-count", text: Message.count, wait: 10)
```
seanpdoyle, line 23: Could we add a test for inline as well?

```ruby
test "New messages update the message count with inline: content" do
  visit messages_path

  assert_text "Messages"
  message.broadcast_update_to(:messages, target: "messages-count",  inline: <<~ERB)
    <%= Message.count %> messages sent
  ERB
  assert_selector("#messages-count", text: Message.count, wait: 10)
end
```

yunggindigo: Yea good point!

test/system/broadcasts_test.rb
```diff
@@ -23,6 +23,16 @@ class BroadcastsTest < ApplicationSystemTestCase
     assert_selector("#messages-count", text: Message.count, wait: 10)
   end
 
+  test "New messages update the message count with inline: content" do
+    visit messages_path
+    
+    message = Message.new(record_id: 1, content: "Message with inline")
+    message.broadcast_update_to :messages, target: "messages-count", 
+      inline: ERB.new("<%= Message.count %> messages sent").result
```
seanpdoyle, line 31: The `inline:` option accepts the ERB _template string_ itself. What this test is covering is essentially the same as the `html:` option.

You can pass the ERB template directly, without the wrapping `ERB.new` and `.result` call. I've edited [my suggestion](https://github.com/hotwired/turbo-rails/pull/245/files/e9c53b80a73f01d004bf216709ad6aafdf5620dc..850c17641f57d01fcc53303c07d6b8056d4f47f2#r714795322) slightly since the first time I posted it, if you want a reference.

yunggindigo: Okay I understand now thanks

yunggindigo: Actually Im thinking maybe the test is just flaky because were testing the same element and using broadcasts to update Ill change the erb test to update a separate element

seanpdoyle: The system tests are flaky, I agree. I've opened https://github.com/hotwired/turbo-rails/pull/248 to try and help with that.

On top of that, your `broadcast_update_to` calls are not wrapped in `assert_appends_text` blocks (which is renamed to `assert_broadcasts_text` in #248), so the underlying broadcast jobs won't execute, so the `<turbo-stream>` elements won't be pushed over Action Cable.

app/models/concerns/turbo/broadcastable.rb
```diff
@@ -296,8 +310,11 @@ def broadcast_rendering_with_defaults(options)
       options.tap do |o|
         # Add the current instance into the locals with the element name (which is the un-namespaced name)
         # as the key. This parallels how the ActionView::ObjectRenderer would create a local variable.
-        o[:locals] = (o[:locals] || {}).reverse_merge!(model_name.element.to_sym => self)
-        o[:partial] ||= to_partial_path
+        # if the html or inline options are passed it will skip setting a partial from #to_partial_path
+        unless o.include?(:html) || o.include?(:inline)
+          o[:locals] = (o[:locals] || {}).reverse_merge!(model_name.element.to_sym => self)
```
seanpdoyle, line 315: Thinking through this further, I believe we'd only want to conditionally omit the `partial:` option when `html:` or `inline:` is present. continuing to accept and forward along `locals:` feels appropriate.

```ruby
o[:locals] = (o[:locals] || {}).reverse_merge!(model_name.element.to_sym => self)

unless o.include?(:html) || o.include?(:inline)
  o[:partial] ||= to_partial_path
end
```

seanpdoyle:
Since you're a first time contributor, a maintainer with GitHub permissions will need to intervene so that your tests to run in CI.

app/models/concerns/turbo/broadcastable.rb
yunggindigo: This is very valid as you can use the locals inside of `inline:` option

test/system/broadcasts_test.rb
seanpdoyle: @yunggindigo if you rebase off `main` and throw away your branch's `Message` changes, these tests should be more predictable. #248 has merged, and `Message` is a full Active Record model.

yunggindigo:
Tests failed last time but pass in development

yunggindigo:
I would really like to get this merged in I need a rerun of the cli @dhh

app/models/concerns/turbo/broadcastable.rb
```diff
@@ -297,7 +311,10 @@ def broadcast_rendering_with_defaults(options)
         # Add the current instance into the locals with the element name (which is the un-namespaced name)
         # as the key. This parallels how the ActionView::ObjectRenderer would create a local variable.
         o[:locals] = (o[:locals] || {}).reverse_merge!(model_name.element.to_sym => self)
-        o[:partial] ||= to_partial_path
+        # if the html or inline options are passed it will skip setting a partial from #to_partial_path
+        unless o.include?(:html) || o.include?(:inline)
```
dhh, line 315: Don't like having both html and inline. Let's just go with `html`.

yunggindigo: Okay sounds good to me I will update it

yunggindigo: @dhh all done

romanos:
Yes these workarounds... work. But it smelled funny so I went with the aforementioned method. I've been keeping my eye out cause I can't be the only one. 

(From mobile)

PR #247, opened 2021-09-23
Add guard to fail CI on uncommitted changes

seanpdoyle:
When pushing to GitHub Actions, fail the test suite if executing
`bin/test` generates changes that are uncommitted.

Similarly, extend `bin/test` to also `yarn install` to ensure that
assets are not out of synchronization.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,3 +33,7 @@ jobs:
         run: |
           bin/rails db:test:prepare
           bin/test test/**/*_test.rb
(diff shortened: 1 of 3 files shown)


dhh:
I don't get this when generating locally. This is removing several commits present in turbo/turbo?

seanpdoyle:
My local yarn dependencies were out of date. I've pivoted this PR to instead add `yarn install` to `bin/test` and to include the git guard in CI.

PR #248, opened 2021-09-23
Promote Dummy app's `Message` to Active Record

seanpdoyle:
In an effort to reduce the number of flaky tests and the amount of
in-memory state leaking between tests, this commit promotes the test
suite's dummy application's `Message` object to an `ApplicationRecord`
descendant.

Something in the test suite configuration is preventing the database
from being wiped between test runs. This results in state leaking
between tests. As a result, our Continuous Integration tests are flaky.

This commit introduces a short-term remedy by adding a `setup {
Message.delete_all }` block to the `test/system/broadcasts_test.rb`.
This should be unnecessary, but until we determine to root cause, it
helps pass the suite more consistently.

### DIFF
--- a/test/system/broadcasts_test.rb
+++ b/test/system/broadcasts_test.rb
@@ -1,12 +1,14 @@
 require "application_system_test_case"
 
 class BroadcastsTest < ApplicationSystemTestCase
+  setup { Message.delete_all }
+
   test "Message broadcasts Turbo Streams" do
     visit messages_path
 
     assert_text "Messages"
-    assert_appends_text "Message 1" do |text|
-      Message.new(record_id: 1, content: text).broadcast_append_to(:messages)
+    assert_broadcasts_text "Message 1" do |text|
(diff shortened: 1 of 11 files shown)


test/system/broadcasts_test.rb
```diff
@@ -5,23 +5,23 @@ class BroadcastsTest < ApplicationSystemTestCase
     visit messages_path
 
     assert_text "Messages"
-    assert_appends_text "Message 1" do |text|
-      Message.new(record_id: 1, content: text).broadcast_append_to(:messages)
+    assert_broadcasts_text "Message 1" do |text|
+      Message.create(id: 1, content: text).broadcast_append_to(:messages)
```
seanpdoyle, line 9: Something is going on in the test suite that prevents the database from being wiped between test runs. This means that state is leaking between tests. This test is failing in CI in spite of the fact that this `Message.create` call is the first and only `Message` record in the test.

PR #250, opened 2021-09-25
Extend ActionDispatch::Request

seanpdoyle:
Implements `Request#turbo_frame?` predicate along with the
`Request#turbo_frame` accessor. Both read from the `Turbo-Frame:`
header.

Mark variant as `turbo_frame` when header present
---

Closes https://github.com/hotwired/turbo-rails/issues/229

Mark the request's [variant][] as `:turbo_frame` whenever the
`Turbo-Frame:` header is present in the request. Also adds documentation
to describe how to achieve the previously built-in layout optimizations.

[variant]: https://guides.rubyonrails.org/4_1_release_notes.html#action-pack-variants


dhh:
I don't love the idea of using variants for turbo frames. We did variants originally to provide different types of HTML responses for different platforms, like a native iOS app. Where what you want to send is something substantially different because of the constraints of the platform. I don't think that applies here.

I'm fine with having a way to detect turbo frame requests as a conditional, but I think that's as far as it should go. This should be the exception. If you're branching all your responses for frames vs not, something isn't right, and we should investigate those pressures in a different way.

Also don't love the idea of Request itself getting extended with this predicate. That's too low level. The integration we're doing for Turbo is at the controller level. I'd rather just keep things there.

WriterZephos:
@dhh I am confused about how to make a controller action respond appropriately to both internal requests that should just replace turbo-frames and requests coming from outside. It seems that layouts get rendered by default, but you would want to turn them off in the former case, and there is no way of doing that conditionally so that you don't mess up the latter case.

Is there a detail or a pattern I am missing here?

arjun810:
> I don't love the idea of using variants for turbo frames. We did variants originally to provide different types of HTML responses for different platforms, like a native iOS app. Where what you want to send is something substantially different because of the constraints of the platform. I don't think that applies here.
> 
> I'm fine with having a way to detect turbo frame requests as a conditional, but I think that's as far as it should go. This should be the exception. If you're branching all your responses for frames vs not, something isn't right, and we should investigate those pressures in a different way.
> 
> Also don't love the idea of Request itself getting extended with this predicate. That's too low level. The integration we're doing for Turbo is at the controller level. I'd rather just keep things there.

Is there a recommended pattern for a layout that has an expensive-to-load sidebar, with a turbo-frame content pane on the right? We need to use turbo_frame_request? as conditionals inline for whether to render the full page, or just the content pane, and it does feel like the variant would be cleaner. Just curious if there's a better pattern out there for this!

PR #254, opened 2021-10-01
Make it easier to opt out of precompilation

ghiculescu:
If you don't want to use use the asset pipeline to precompile `turbo.js`, it is not clear how to opt out at the moment. This makes it a bit easier to by introducing a const. I copied the pattern in [sprockets](https://github.com/rails/sprockets-rails/blob/118ce60b1ffeb7a85640661b014cd2ee3c4e3e56/lib/sprockets/railtie.rb#L85.

A reason why you might want this is if you're using webpack to build your JS (but still want the `turbo-rails` gem for the server side helpers it adds).

I think a configuration option for this would be even better, but the gem has no other config so far and I wasn't sure if this was a good enough reason to break that.


ghiculescu:
Test seems intermittent / related.

Issue #243, opened 2021-09-22
Turbo stream partial with button_to tag throws ActionView::Template::Error


pettersj:
`after_create_commit` with `broadcast_append_to` throws this error:
`ActionView::Template::Error (Request forgery protection requires a working session store but your application has sessions disabled. You need to either disable request forgery protection, or configure a working session store.)
`
(seems to happen to all broadcasts with partials) 

Log trace:
```
10:22:02 web.1  | ActionView::Template::Error (Request forgery protection requires a working session store but your application has sessions disabled. You need to either disable request forgery protection, or configure a working session store.):
10:22:02 web.1  |     13:     <div class="flex">
10:22:02 web.1  |     14: 
10:22:02 web.1  |     15:       <div class="up">
10:22:02 web.1  |     16:         <%= button_to increment_question_group_path(question_group), method: :patch do %>
10:22:02 web.1  |     17:           <svg xmlns="http://www.w3.org/2000/svg" class="h-5 w-5 text-gray-500 font-light" viewBox="0 0 20 20" fill="currentColor">
10:22:02 web.1  |     18:             <path fill-rule="evenodd" d="M5.293 9.707a1 1 0 010-1.414l4-4a1 1 0 011.414 0l4 4a1 1 0 01-1.414 1.414L11 7.414V15a1 1 0 11-2 0V7.414L6.707 9.707a1 1 0 01-1.414 0z" clip-rule="evenodd" />
10:22:02 web.1  |     19:           </svg>
10:22:02 web.1  |   
10:22:02 web.1  | app/views/question_groups/_question_group.html.erb:16
10:22:02 web.1  | app/models/question_group.rb:16:in `append_to_list'
```

Rails 7 alpha2, turbo-rails 0.7.14

pettersj:
The same is happening with forms that are either post, patch or delete. method: :get works fine.

dhh:
Are you sure you have a working session store, as the error mentions? Can you create a boiled down app that replicates this? Not seeing it on my end.

pettersj:
Yeah, the session store is fine. There's something strange going on here, because a it actually does replace with a normal _form.html.erb. I'll close it for now, won't bother you with it, I'm definitely something wrong 😅

DavidColby:
I believe this needs to be reopened.

Rails 7 alpha [raises an error](https://github.com/rails/rails/pull/42231) when sessions are disabled, which appears to break model broadcasts.

[...]

PR #256, opened 2021-10-11
Disable CSRF when rendering from turbo

casperisfine:
Ref: https://github.com/hotwired/turbo-rails/issues/243

CSRF protection requires access to the user session, but it is only available during a regular request cycle, not when rendering from the background.

I think the solution should be something like that.


casperisfine:
Closing in favor of https://github.com/rails/rails/pull/43427

PR #272, opened 2021-11-10
Slice '#' off target start if present

### LINKED ISSUE #146: using dom_id as a stream target
given a frame target 
```ruby
<%= turbo_frame_tag dom_id(order, :some_prefix) do %>
...
<% end %>
```

```ruby
     broadcast_replace_to [order, :order_index_invoice_stream],
                          target: :order_index_invoice_frame,
                          target: dom_id(order, :some_prefix),
                           partial: 'orders/index/seller/shared/invoice_button',
                           locals: { order: order }
```

the target will stream back over ws with target: '#some_prefix_order_123' whereas it seems turbo expects the target to not have a prepended '#' and expects 'some_prefix_order_123' in order the match the frame id

julianrubisch:
Eventually fixes #146

please let me know if I missed something and/or there's a better place to reference this from

### DIFF
--- a/app/helpers/turbo/streams/action_helper.rb
+++ b/app/helpers/turbo/streams/action_helper.rb
@@ -26,7 +26,7 @@ module Turbo::Streams::ActionHelper
       if target.respond_to?(:to_key)
         [ ("#" if include_selector), ActionView::RecordIdentifier.dom_id(target) ].compact.join
       else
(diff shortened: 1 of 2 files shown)


seanpdoyle:
Could we add a test to exercise the behavior and to guard against regressions in the future?

PR #275, opened 2021-11-14
Edit error message for clarity

pitosalas:
See https://github.com/hotwired/hotwire-rails/issues/45


dhh:
Need different paths. Because you could get this error if you're on Rails 7 too.

PR #279, opened 2021-12-05
Update generated `turbo.min.js.map`

seanpdoyle:
The contents of this commit were generated by executing `bin/test`
locally.


seanpdoyle:
The current version of `main` (without these changes) returns `1` as an exit code for both `git update-index --refresh` and `git diff-index --quiet HEAD --` commands executed in the [CI guard](https://github.com/hotwired/turbo-rails/blob/6958e0e2742d4744da80128d178e7255989386a4/.github/workflows/ci.yml#L44-L45) against this. I'm not sure how that happened.

dhh:
Diff looks weird?

Screen Shot 2021-12-06 at 09 15 25

seanpdoyle:
These changes were generated due to some quirks in my local setup, namely a `yarn link @hotwired/turbo` call to reference a local branch in a local clone of `hotwired/turbo`. The version of the source maps on `main` is both correct and synchronized with the output of `bin/test`. Closing this.

Issue #171, opened 2021-05-05
What is the purpose of Turbo::Broadcastable::broadcasts ?


fig:
From my reading of the [documentation](https://github.com/hotwired/turbo-rails/blob/2945dcb7c5fc634ad7d459e0872cf04cf8a3576d/app/models/concerns/turbo/broadcastable.rb#L60), the class method `broadcasts` should configure the model to broadcast creates, updates, and destroys to a stream name automatically set to the current model. I would expect the broadcasts to be made to a stream name derived from the model name pluralised, although the pluralisation is not documented explicitly.

In any case, the actual implementation broadcasts to a stream name derived from the `GlobalID` of the model instance. This makes little sense because although a partial can be configured to accept broadcasts from this unique stream name, updates will replace the target element as expected, destroys will leave an  orphaned `turbo-cable-stream-source` tag behind and creating a new model instance has no effect because the broadcast will be made to a channel name that does not exist at the time of rendering.

What are community's thoughts on this?

Apologies if this is the wrong place to ask.

danmcge:
Let's take this to https://discuss.hotwire.dev/.

(Briefly, though, I think what you’re describing is, say, `broadcasts_to :posts`. That’s not what `broadcasts` is for. Imagine you had a `post` that had comments that `broadcasts_to :posts`, but it’s important, for some reason, to broadcast when the title of the post changes. That’s when you’d reach for `broadcasts`. The `<%= turbo_stream_for @post %>` call in your template would pick up both the comments and the change to the post itself. And it would only be the changes to, and the comments for, that specific post, because the stream is scoped by the ID.)

PR #295, opened 2022-01-04
Turbo::Broadcastable#broadcasts broadcasts creates to model_name.plural stream

Exterm1nate:
As mentioned in https://github.com/hotwired/turbo-rails/issues/171, `Turbo::Broadcastable#broadcasts` provides the stream for broadcasting creates that can't be on the page until the model instance is created.
The issue was closed without any conversation and I didn't find any topics at Hotwire forum about this problem.
My solution is to broadcast instance creation to stream with model plural name, which looks quite intuitive for me.

My apologies if i missed something.

### DIFF
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -71,9 +71,10 @@ module Turbo::Broadcastable
       after_destroy_commit -> { broadcast_remove_to stream.try(:call, self) || send(stream) }
     end
 
-    # Same as <tt>#broadcasts_to</tt>, but the designated stream is automatically set to the current model.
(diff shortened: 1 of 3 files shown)


rights-uchida (approved):

fig:
Hey @Exterm1nate , Thanks for picking this up! I was he author of #171.

I didn't really know at the time whether I had a valid point, or I just didn't understand the `Broadcastable` module.

I'm happy that someone else has confirmed my suspicions and delighted that you have proposed a patch. :+1:

dhh:
Actually, I still don't fully understand why if you need an overwritten stream name that you don't just use `broadcasts_to`? That's explicitly what that's there for?

dhh:
Ah, never mind. Finally see the problem when it was shown in #312.

PR #298, opened 2022-01-14
Add Turbo::Broadcastable support for #broadcast_render and #broadcast_render_to

morgangrubb:
This is very useful for when a destroy action contains a lot of functionality.

`app/views/messages/_destroy.turbo_stream.erb`
```erb
<%= turbo_stream.replace(dom_id(message, :modal_body)) do %>
  <%= render "messages/body", message: message %>
<% end %>

<%= turbo_stream.remove(dom_id(message, :outer)) %>

<% if message.item.messages.none? %>
  <%= turbo_stream.remove(dom_id(message.item, :messages)) %>
<% end %>
```

`app/models/message.rb`
```ruby
  after_destroy_commit {
    broadcast_render partial: "messages/destroy"
  }
end
```

Right now it's possible but I have to call `Turbo::StreamsChannel` directly:
```ruby
  after_destroy_commit {
    Turbo::StreamsChannel.broadcast_render **broadcast_rendering_with_defaults(partial: "messages/destroy")
  }
end
```

### DIFF
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -45,8 +45,9 @@
 # within a real-time path, like a controller or model, since all those updates require a rendering step, which can slow down
 # execution. You don't need to do this for remove, since only the dom id for the model is used.
 #
-# In addition to the four basic actions, you can also use <tt>broadcast_render_later</tt> or
-# <tt>broadcast_render_later_to</tt> to render a turbo stream template with multiple actions.
+# In addition to the four basic actions, you can also use <tt>broadcast_render</tt>,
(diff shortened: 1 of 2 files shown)


dhh:
If you're rendering templates, you really shouldn't do it inline. We have `broadcast_render_later` for this to ensure it's out of band. Is there are particular reason you want to do it sync?

morgangrubb:
In this particular case it's because there are a few different clean up operations that should happen when the item gets deleted (`_destroy.turbo_stream.erb`) and they can't be handled by a `_later` operation owing to the item no longer being in the database when the job runs.

dhh:
I see. The reason we normally run remove in-band is because there's no rendering going on. So maybe we can just document these trade-offs under the later method? Basically: You shouldn't render in-band under normal circumstances. If you really know what you're doing, fine, here's a sharp knife.

Issue #299, opened 2022-01-22
turbo:install:redis redis_installed? doesn't work on Windows


mohits:
Installing rails and doing a `rails new` on Windows works perfectly well with one warning/ error that comes up when the rails new is running.

> The system cannot find the path specified.

I have been able to track it down to `redis_installed?` at: https://github.com/hotwired/turbo-rails/blob/acbb9250cf97cbf9f00570cbfbdeeffc815003b8/lib/tasks/turbo_tasks.rake#L6

The output is redirected to `dev/null` causes Windows to output that line (and of course, even if that was not the case `which` is not available on Windows).

As @dhh noted in https://github.com/hotwired/turbo-rails/issues/225#issuecomment-906490561 "it doesn't have to work perfectly everywhere" and I can confirm it does not :)

Just raising it here. Also, I have [documented this on my post about Rails 7 + Windows](https://notepad.onghu.com/2021/rails7-on-windows-it-just-works/#upd1) that people don't need to worry too much about whether Rails was set up properly when they see that message.

dhh:
Would be happy to take a PR that makes this work better on Windows.

mohits:
Thanks @dhh.

Step 1 was identification.
Step 2 is finding the solution. Based on what I see and know, this will be closest to the correct solution. This also searches for redis-server using where instead of which, and redirects correctly to NUL for both stdout and stderr.

```ruby
def redis_installed?
  Gem.win_platform? ?
    system('where redis-server > NUL 2>&1') : 
    system('which redis-server > /dev/null')
end
```

I have done some quick checks with this on Windows (CRuby and JRuby) using both Powershell and Command Prompt and it looks good. I will do a couple of checks on my Ubuntu VM later this week and then get into the process of the PR.

mohits:
#300 is merged.
This is now closed.

PR #300, opened 2022-01-25
Use where on Windows instead of which (#299)

mohits:
If the call is on Windows, it uses `where` instead of `which` and correctly sends the output to NUL so that it does not show unnecessary message on the installer's output. This fixes the incorrect/ misleading message shown on Windows currently.


mohits:
Verified that it works fine on Windows with JRuby/ CRuby with both Powershell and Command (system uses 'command' anyway). Also, verified that it works as expected on Ubuntu with and without redis-server being present.

PR #269, opened 2021-10-29
Problem: inability to pass additional parameters to stream channels

yrashk:
Recently, an ability to customize channel has been added (https://github.com/hotwired/turbo-rails/pull/264).

However, it's still quite rigid as we can't pass any contextual information
from the client side. This can be very useful when, for example, there's a need
to indicate state of the page for the initial burst of messages upon
subscription.

Solution: allow passing additional parameters through `data-` attributes

This integrates nicely with `turbo_stream_from` helper which will be able
to process `data: {...}` attributes to pass them to the custom element.

`turbo-cable-stream-source` implementation has been updated to take these
dataset attributes into account and pass them along upon subscription.

PR #308, opened 2022-02-16
Problem: inability to pass additional parameters to stream channels

yrashk:
Recently, an ability to customize a channel has been added (https://github.com/hotwired/turbo-rails/pull/264).

However, it's still quite rigid as we can't pass any contextual information
from the client-side. This can be very useful when, for example, there's a need
to indicate the state of the page for the initial burst of messages upon
subscription.

Solution: allow passing additional parameters through `data-` attributes

This integrates nicely with `turbo_stream_from` helper which will be able
to process `data: {...}` attributes to pass them to the custom element.

`turbo-cable-stream-source` implementation has been updated to take these
dataset attributes into account and pass them along upon subscription.

---

Supersedes #269


yrashk:
CI failures seem to be unrelated to the change itself. I will happily accept any guidance to get them bent back in shape ;)

PR #323, opened 2022-04-08
Add `append_body` and `prepend_body` methods to TagBuilder

swanson:
Sometimes you just want to add content "anywhere" on the page:
- Common for adding modals/tooltips with absolute/fixed position
- Stimulus controllers that run one-off functions (e.g. [callout_controller](https://github.com/hotwired/stimulus-site/blob/35a4d71754278ae5ecdfc51fedd06b87a28b285b/_source/reference/lifecycle_callbacks.md?plain=1#L10) from Stimulus website, [Beacons](https://developer.mozilla.org/en-US/docs/Web/API/Beacon_API), etc)
- Adding extra `<script>` tags for 3rd party tools

The Rails helpers supports this, but you would need to use `prepend_all` so you could pass in "body" as the query selector (`prepend "body"` would try to look for an element with `id=body`). This doesn't read well since you won't have multiple `<body>` tags.

This PR adds helpers (`append_body` and `prepend_body`) for appending or prepending content to the `<body>` tag of the document directly.

image

### DIFF
--- a/app/models/turbo/streams/tag_builder.rb
+++ b/app/models/turbo/streams/tag_builder.rb
@@ -224,7 +224,7 @@ class Turbo::Streams::TagBuilder
     action_all :prepend, targets, content, **rendering, &block
   end
 
-  # Preped to the <body> tag of the document either the <tt>content</tt> passed in or a
+  # Prepend to the <body> tag of the document either the <tt>content</tt> passed in or a
   # rendering result determined by the <tt>rendering</tt> keyword arguments, the content in the block,
   # or the rendering of the content as a record. Examples:
   #


swanson:
@dhh any thoughts on this?

dhh:
I think the pattern of just appending to the body is solid, but I don't think the sugar-coated API adds enough to offset the complication of expanding the surface area. 

I'd also suggest just making it clearer in your own DOM by having an empty container as part of your layout with the id of `beacons`, since you can then do `turbo_stream.append_to "beacons" do ...`.

swanson:
@dhh any reconsiderations for this now that 7.2 will have the ability to respond to GETs with turbo streams? Seems like many people will start using that to append modals into the page.

I think a container with `id=modals` is still a fine approach but thought I'd bump this given the new changes. Thanks!

dhh:
The trade off hasn't fundamentally changed for me. I think there's a lot of value in keeping the main set of turbo stream actions low in other to have an approachable, easy to learn API. I don't think the additional sugar of going from `append_all "body"` to `append_body` is worth it. But of course, it's pretty easy to freedom-patch your own setup to have this.

PR #324, opened 2022-04-12
Offer a flag to opt-out of native routes

alexandreruban:
In this PR, we add the `config.turbo.draw_routes` configuration option to opt-out of
native routes.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,6 +1,7 @@
-# FIXME: Offer flag to opt out of these native routes
+# frozen_string_literal: true
+
 Rails.application.routes.draw do
   get "recede_historical_location"  => "turbo/native/navigation#recede",  as: :turbo_recede_historical_location
   get "resume_historical_location"  => "turbo/native/navigation#resume",  as: :turbo_resume_historical_location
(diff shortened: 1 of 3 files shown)


tonysm:
on that note, is there any documentation anywhere on how to use these routes? @alexandreruban 😅

dhh:
What sparked this? How are these routes interfering?

alexandreruban:
When typing `rails routes`, those routes appear even if we don't plan to use Turbo Native.

With other engines such as `ActiveStorage` or `ActionMailer`, if we don't plan to use them, we can simply remove the engine to remove the routes. However, with `turbo-rails`, we might want to use Turbo Frames and Turbo Streams but not Turbo Native.

dhh:
Okay. I think this is kinda borderline. The routes, even if unused, don't pose any complication or serious performance issue. Whereas every option we add complicate the overall package. 

What I'd rather actually see would be a way where you could delete named routes in Rails. Basically something like `erase :turbo_recede_historical_location`. That would just remove that named route from the routes. Then you could remove anything that was added you wouldn't like to have.

Another alternative would be to have something like `rake routes:app`, perhaps, which would only display the routes added by your app (and not by the framework).

Feel free to explore either of those options as Rails PRs!

PR #327, opened 2022-05-03
Improve upon test suite flakiness

seanpdoyle:
Make `turbo_test.rb` with Rails' generated `test_helper.rb`
===

> Something in the test suite configuration is preventing the database
> from being wiped between test runs. This results in state leaking
> between tests. As a result, our Continuous Integration tests are flaky.
>
> - [hotwired/turbo-rails#248][]

As a follow-up to the [short-term solution][] shipped in
[hotwired/turbo-rails#248][], this commit attempts to make the
`test/turbo_test.rb` file's setup consistent with the test harness setup
generated by Rails' [engine generator][] code.

To that end, this commit:

* renames the `test/turbo_test.rb` file to `test/test_helper.rb`
* omits one-off `require` calls for particular dependencies
* re-orders the require calls so that the
  `../test/dummy/config/environment` file is required ahead of the
  `rails/test_help` file

[engine generator]: https://github.com/rails/rails/blob/3c48b4030adbded21bebaa0d78254216cca48a6e/railties/lib/rails/generators/rails/plugin/templates/test/test_helper.rb.tt
[hotwired/turbo-rails#248]: https://github.com/hotwired/turbo-rails/pull/248
[short-term solution]: https://github.com/hotwired/turbo-rails/commit/c2dc5b15acaf3cb0aebb581b7dfa0b792ec50bc1

Use Ruby 2.7 argument forwarding
===

Switching to argument forwarding addresses deprecation warnings like:

```
hotwired/turbo-rails/test/streams/streams_channel_test.rb:140: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
hotwired/turbo-rails/test/turbo_test.rb:14: warning: The called method `render' is defined here
```

Tests: Load 6.1 defaults in Dummy Application
===

Resolve deprecation warnings like:

```
Preparing test database
DEPRECATION WARNING: Non-URL-safe CSRF tokens are deprecated. Use 6.1 defaults or above. (called from <top (required)> at /home/runner/work/turbo-rails/turbo-rails/test/dummy/config/initializers/inspect_helpers.rb:1)
DEPRECATION WARNING: Using legacy connection handling is deprecated. Please set
`legacy_connection_handling` to `false` in your application.

The new connection handling does not support `connection_handlers`
getter and setter.

Read more about how to migrate at: https://guides.rubyonrails.org/active_record_multiple_databases.html#migrate-to-the-new-connection-handling
 (called from <top (required)> at /home/runner/work/turbo-rails/turbo-rails/test/test_helper.rb:6)
```

Since our GitHub CI matrix includes `6.1`, `7.0`, and `main`, CI's tests
should run with at least the `6.1` defaults.

CI: Continue other executions on error
===

Remove the `continue-on-error` configuration and instead allow other
jobs complete in spite of failures.

Improve flaky Broadcast tests
===

First, don't render HTML with the `<turbo-cable-stream-source>` element.
Instead, append the element when clicking a `<button>`.

Next, remove all waiting from Capybara tests, and rely on default
values.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,6 @@ jobs:
           - "6.1"
           - "7.0"
           - "main"
-        continue-on-error: [ false ]
 
         # Disabled until minitest relaxes its upper bound: https://github.com/seattlerb/minitest/pull/862
         # > minitest-5.14.2 requires ruby version < 3.1, >= 2.2, which is incompatible with the current version, ruby 3.1.0p-1
@@ -25,7 +24,7 @@ jobs:
 
     name: ${{ format('Tests (Ruby {0}, Rails {1})', matrix.ruby-version, matrix.rails-version) }}
     runs-on: ubuntu-latest
-    continue-on-error: ${{ matrix.continue-on-error }}
+    continue-on-error: true
 
     steps:
       - uses: actions/checkout@v1
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -1,4 +1,4 @@
-require "turbo_test"
+require "test_helper"
 
 class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   driven_by :selenium, using: :headless_chrome, screen_size: [1400, 1400]
--- a/test/drive/drive_helper_test.rb
+++ b/test/drive/drive_helper_test.rb
@@ -1,4 +1,4 @@
-require "turbo_test"
+require "test_helper"
 
 class Turbo::DriveHelperTest < ActionDispatch::IntegrationTest
   test "opting out of the default cache" do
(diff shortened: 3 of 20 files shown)


seanpdoyle:
The current batch of CI failures here (and in https://github.com/hotwired/turbo-rails/pull/328) seem related to [test/system/broadcasts_test.rb](https://github.com/hotwired/turbo-rails/blob/d4e26fec0ed02e06cd3ab26882dc234fccd2972a/test/system/broadcasts_test.rb#L40-L45), which was recently introduced in https://github.com/hotwired/turbo-rails/pull/308.

dhh:
Cc @yrashk

dhh:
Seems like we still have some Rails 6.1 failures?

seanpdoyle:
This is passing, finally.

PR #328, opened 2022-05-03
Make `turbo_test.rb` consistent with Rails' generated `test_helper.rb`

seanpdoyle:
I've opened this as a draft PR, since I'm unsure about the root-cause of database records leaking across test runs
---

> Something in the test suite configuration is preventing the database
> from being wiped between test runs. This results in state leaking
> between tests. As a result, our Continuous Integration tests are flaky.
>
> -- [hotwired/turbo-rails#248][]

As a follow-up to the [short-term solution][] shipped in
[hotwired/turbo-rails#248][], this commit attempts to make the
`test/turbo_test.rb` file's setup consistent with the test harness setup
generated by Rails' [engine generator][] code.

To that end, this commit:

* renames the `test/turbo_test.rb` file to `test/test_helper.rb`
* omits one-off `require` calls for particular dependencies
* re-orders the require calls so that the
  `../test/dummy/config/environment` file is required ahead of the
  `rails/test_help` file

[engine generator]: https://github.com/rails/rails/blob/3c48b4030adbded21bebaa0d78254216cca48a6e/railties/lib/rails/generators/rails/plugin/templates/test/test_helper.rb.tt
[hotwired/turbo-rails#248]: https://github.com/hotwired/turbo-rails/pull/248
[short-term solution]: https://github.com/hotwired/turbo-rails/commit/c2dc5b15acaf3cb0aebb581b7dfa0b792ec50bc1

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -1,4 +1,4 @@
-require "turbo_test"
+require "test_helper"
 
 class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   driven_by :selenium, using: :headless_chrome, screen_size: [1400, 1400]
--- a/test/drive/drive_helper_test.rb
+++ b/test/drive/drive_helper_test.rb
@@ -1,4 +1,4 @@
-require "turbo_test"
+require "test_helper"
 
 class Turbo::DriveHelperTest < ActionDispatch::IntegrationTest
   test "opting out of the default cache" do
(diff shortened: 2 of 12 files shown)


seanpdoyle:
Unfortunately, this didn't resolve the [CI failures](https://github.com/hotwired/turbo-rails/runs/6275835813?check_suite_focus=true#step:4:25).

My hunch is that there's a missing or misconfigured Active Record value somewhere that's resulting in records persisting across test runs.

seanpdoyle:
Closing in favor of unifying streams of work into https://github.com/hotwired/turbo-rails/pull/327.

Issue #330, opened 2022-05-04
active_job must be required


aaronjensen:
Because of https://github.com/hotwired/turbo-rails/blob/main/app/jobs/turbo/streams/broadcast_job.rb and https://github.com/hotwired/turbo-rails/blob/main/app/jobs/turbo/streams/action_broadcast_job.rb and the eager loader, `active_job` needs to be required by any app that is using `turbo-rails`. 

Should `turbo-rails` somehow make these optional or require `active_job` explicitly?

A user who does not require `active_job` will see an `uninitialized constant ActiveJob` typically when they deploy to a production-like environment for the first time.

PR #331, opened 2022-05-05
Fix missing `activejob` dependency

codealchemy:
closes https://github.com/hotwired/turbo-rails/issues/330

Removed in https://github.com/hotwired/turbo-rails/pull/283/commits/aea2fecafdb304c2c55820bae456afdf729f75d3 (which this effectively reverts, that change was a part of https://github.com/hotwired/turbo-rails/pull/283), `ActiveJob` is used in a couple areas https://github.com/hotwired/turbo-rails/blob/b9b87fc627d9148899970deea01069875f15db3f/app/jobs/turbo/streams/broadcast_job.rb#L3 and https://github.com/hotwired/turbo-rails/blob/b9b87fc627d9148899970deea01069875f15db3f/app/jobs/turbo/streams/action_broadcast_job.rb#L2

### DIFF
--- a/turbo-rails.gemspec
+++ b/turbo-rails.gemspec
@@ -11,6 +11,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = ">= 2.6.0"
 
+  s.add_dependency "activejob", ">= 6.0.0"
   s.add_dependency "actionpack", ">= 6.0.0"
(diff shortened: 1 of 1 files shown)


aaronjensen:
Is this enough? Doesn’t it need to be required?

Issue #336, opened 2022-05-23
Module parse failed: Unexpected token when transpiling with webpack due to inclusion of TypeScript optional chaining operator (`submitter?.formMethod`)


davidrunger:
I am attempting to bump @hotwired/turbo-rails from 7.1.1 to 7.1.2, but I am getting this error when running `bin/webpack` in my Rails app:

```
ERROR in ./node_modules/@hotwired/turbo-rails/app/javascript/turbo/form_submissions.js 9:31
Module parse failed: Unexpected token (9:31)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|   }
| }) {
>   const formMethod = submitter?.formMethod;
|
|   if (formMethod && fetchRequest.body.has("_method")) {
 @ ./node_modules/@hotwired/turbo-rails/app/javascript/turbo/index.js 2:0-66 7:39-67
 @ ./app/javascript/packs/quizzes.js
```

It seems that the error is being caused by the inclusion of the TypeScript optional chaining operator (`?.`) in that file (`app/javascript/turbo/form_submissions.js`) in `submitter?.formMethod`.

I think that this issue was introduced in https://github.com/hotwired/turbo-rails/pull/239 (Support `[formmethod]` overrides to `_method`) here: https://github.com/hotwired/turbo-rails/pull/239/files#diff-f438134433fd9897751d1b5cd219ea33ac29be8f3af3e4ac17e65475e4cadbeeR2.

dhh:
The chaining operator should be supported from Node 14+. I think we should probably still address this, but curious what version you're on? Maybe we need to find a way to have version requirements.

davidrunger:
Maybe I should also mention that I am using `webpacker` (version 5.4.3), which I know has been retired. I'm not sure if that might be a factor.

davidrunger:
Inspired by [this Stack Overflow answer](https://stackoverflow.com/a/57689106/4009384), I am able to get webpack to compile successfully by adding the following to my webpack configuration (which, for my app, is located in `config/webpack/shared.js`):

```js
// ...
module.exports = {
  // ...
  module: {
    rules: [
      {
        test: /\.js$/,
        loader: 'babel-loader',
        options: {
          plugins: [
            '@babel/plugin-proposal-optional-chaining',
          ],
        },
      },
      // ...
    ],
  },
  // ...
};
```

[...]

PR #337, opened 2022-05-23
Replace `?` operator with conditional check

seanpdoyle:
Closes [#336][]

Address downstream Rails consumers' webpack syntax issues by replacing
the `?` operator with a check that `submitter` is present.

[#336]: https://github.com/hotwired/turbo-rails/issues/336


davidrunger:
Thanks! I have confirmed that this fixes the specific problem that I reported in #336. 👍 When I install `@hotwired/turbo-rails` via `yarn add https://github.com/seanpdoyle/turbo-rails\#inline-conditional-chaining, `bin/webpack` now completes for me without any errors.

However, now that webpack is compiling and so I can run the test suite, I am seeing a newly failing feature test in the part of my app that uses turbo. I haven't yet had a chance to try to figure out what exactly is causing that test failure, though I plan to look into it ASAP. Presumably the test failure has something to do with the changes introduced in @hotwired/turbo-rails@7.1.2 / https://github.com/hotwired/turbo-rails/pull/239, but I don't yet know if it's something I am doing wrong on my end that is incompatible with @hotwired/turbo-rails@7.1.2 vs. a bug or unintended breaking change that might have been introduced into @hotwired/turbo-rails with that code change.

davidrunger:
Scratch what I said above about a newly failing test in my test suite; I think it was just a flaky test. Upon re-running the test suite locally and running the test suite on my CI server, it passes. So this change seems good/fine to me! Thanks again.

dhh:
Released as 7.1.3 on npm 👍

glaszig:
can i get a 1.1.2 of the gem as well please? `?.` is tripping up my safari 12 as well :(

Issue #178, opened 2021-05-25
Allowing usage of layout for turbo stream rendering


rodrigoruas:
Hello,

It seems currently not possible to use layouts for **turbostream views rendering** and it may be a nice addition to deal with cases like flash updates.

Example: `application.turbo_stream.erb`

```ruby 
<turbo-stream action="append" target="flash">
  <template>
    <%= render partial: 'layouts/flash' %>
  </template>
</turbo-stream>
<%= yield %>
```

To make the layout above works, we had to force it:
 ```ruby 
respond_to do |format|
      format.turbo_stream do
        render layout: 'application'
      end
   end
```

As @jduff [pointed out to me](https://github.com/hotwired/turbo/issues/273),  layout is always false when it's a turbo_stream request:

https://github.com/hotwired/turbo-rails/blob/2945dcb7c5fc634ad7d459e0872cf04cf8a3576d/app/controllers/turbo/frames/frame_request.rb#L16

For **turbostream inline rendering** we are using the following approach:
 ```ruby 
def turbo_flash
    turbo_stream.append('flash', partial: 'layouts/flash')
 end
```

```ruby
render turbo_stream: turbo_flash + turbo_stream.update(.....)
          
```

As DHH suggested, we shouldn't do multiple inline turbostreams and we are looking for some better alternatives like the first example above. Do you think forcing the layout is the correct approach or we can have another option?
Thanks!

WriterZephos:
I managed this by rendering the layout as the template and passing in a local value which is the actual template I wanted to render, and the layout rendered that internally where it normally would yield.

So render a layout where instead of 

`<%= yield %>`

you have

`<%= render template %>`

and pass the template var as a local (or you can use `@template` and set it in the controller).

I agree there should be an easier way though.

Octobread:
I was able to achieve this by setting `layout -> { nil if turbo_frame_request? }` in my `ApplicationController`. It's a quick-fire way to enable this by default and seems to work so far (at least for my project).

alextakitani:
I really think that having to manually set flash messages when using Hotwire is a step back from what we had with "regular" Rails.

Using the layout brings it back, why this is not the default behavior? 

Does this cause any problems?

dhh:
You can use layouts with frames by overwriting the layout prescription set in the FrameRequest concern: https://github.com/hotwired/turbo-rails/blob/main/app/controllers/turbo/frames/frame_request.rb#L16.

[...]

Issue #25, opened 2020-12-24
Allow to render stream responses with a layout


Intrepidd:
It would be fantastic to make `render turbo_stream:` use an optional layout

This way, we can have global-level stream updates that make sense on every page.

For instance, our layout could look like this : 

```erb
<% if flash[:success].present? %>
  <%= turbo_stream.replace("flash_success") do %>
    <%= render "flash/success" %>
  <% end %>
<% end>
<%= yield %>
```

Allowing our controller to use flash regardless of the format and then have the layout pick it up either in the stream or in the template.

WDYT ? Happy to provide a PR

dhh:
Pretty interesting idea. Would like to see what an implementation could look like without getting weird 👍

Intrepidd:
I did this very naive approach for now : 

https://github.com/hotwired/turbo-rails/compare/main...Intrepidd:feature/layout

Some pain points : 
- Can't use `options[:layout]` as it already contains informations about the layout used in the controller
- Can't properly use `yield` in the layout code and had to use regular locals instead. Didn't find a way in rails to render a template with a layout when the template is actually a string of already rendered HTML

I probably missed a lot of things,  if someone gives me pointers on a better way to achieve this I'll be happy to look at it.

dhh:
We'd have to solve at least yield for this to be a tenable path. Feel free to open a PR if you manage that.

seanpdoyle:
@Intrepidd is the proposed change scoped to rendering _from_ a controller? Something like

```ruby
def create
  respond_to do |format|
    format.turbo_stream { render turbo_stream: ..., layout: ... }
  end
end
```

Since Turbo Streams are their own MIME type to Rails, the typical layout rendering system should suffice.

I've relied on a layout for this exact use case to render the [FlashHash#alert](https://edgeapi.rubyonrails.org/classes/ActionDispatch/Flash/FlashHash.html#method-i-alert) and [FlashHash#notice](https://edgeapi.rubyonrails.org/classes/ActionDispatch/Flash/FlashHash.html#method-i-notice):

```erb
<%# app/views/layouts/application.turbo_stream.erb %>

<% if alert.present? %>
  <%= turbo_stream.append "flash_messages" do %>
    <%= flash_tag alert, type: :alert %>
  <% end %>
<% end %>

<% if notice.present? %>
  <%= turbo_stream.append "flash_messages" do %>
    <%= flash_tag notice, type: :notice %>
  <% end %>
<% end %>

[...]

PR #351, opened 2022-06-26
Document how to add a layout for turbo frame requests

alexandreruban:
References [#178](https://github.com/hotwired/turbo-rails/issues/178#issuecomment-1159654741)

Adding a layout for `turbo frame` requests is a very common feature that is currently not documented. This PR documents how to add a layout for `turbo frame` requests with the common example of flash messages.

### DIFF
--- a/app/controllers/turbo/frames/frame_request.rb
+++ b/app/controllers/turbo/frames/frame_request.rb
@@ -9,6 +9,20 @@
 # Turbo Frames knows how to fish out the relevant frame regardless.
 #
 # This module is automatically included in <tt>ActionController::Base</tt>.
+#
+# You can override this behavior in <tt>ApplicationController</tt> if you want to have a layout for turbo frame requests:
+#
+#   # app/controllers/articles_controller.rb
+#   class ApplicationController < ActionController::Base
+#     layout -> { "application" if turbo_frame_request? }
+#   end
+#
+# Then, you have to create a layout with the corresponding name. It can be handy for flash messages, for example:
+#
+#   # app/views/layouts/application.turbo_stream.erb
+#   <%= turbo_stream.update('flash', partial: 'layouts/flash') %>
+#
+#   <%= yield %>
 module Turbo::Frames::FrameRequest
   extend ActiveSupport::Concern
 
--- a/app/controllers/turbo/requests.rb
+++ b/app/controllers/turbo/requests.rb
@@ -10,11 +10,11 @@
 #
 # This module is automatically included in <tt>ActionController::Base</tt>.
 #
-# You can override this behavior in <tt>ApplicationController</tt> if you want to have a layout for turbo frame requests:
+# You can override this behavior in <tt>ApplicationController</tt> if you want to have a layout for turbo stream requests:
 #
 #   # app/controllers/articles_controller.rb
 #   class ApplicationController < ActionController::Base
-#     layout -> { "application" if turbo_frame_request? }
+#     layout -> { (turbo_frame_request? && !turbo_stream_request?) ? false : "application" }
 #   end
 #
 # Then, you have to create a layout with the corresponding name. It can be handy for flash messages, for example:
@@ -23,7 +23,7 @@
 #   <%= turbo_stream.update('flash', partial: 'layouts/flash') %>
 #
 #   <%= yield %>
-module Turbo::Frames::FrameRequest
+module Turbo::Requests
   extend ActiveSupport::Concern
 
   included do
@@ -34,6 +34,10 @@ module Turbo::Frames::FrameRequest
   private
     def turbo_frame_request?
       turbo_frame_request_id.present?
+    end
+
+    def turbo_stream_request?
+      request.accept.starts_with?(Mime[:turbo_stream])
     end
 
     def turbo_frame_request_id
--- a/lib/turbo/engine.rb
+++ b/lib/turbo/engine.rb
@@ -36,7 +36,7 @@ module Turbo
 
     initializer "turbo.helpers", before: :load_config_initializers do
       ActiveSupport.on_load(:action_controller_base) do
-        include Turbo::Streams::TurboStreamsTagBuilder, Turbo::Frames::FrameRequest, Turbo::Native::Navigation
+        include Turbo::Requests, Turbo::Streams::TurboStreamsTagBuilder, Turbo::Native::Navigation
         helper Turbo::Engine.helpers
       end
     end


app/controllers/turbo/frames/frame_request.rb
```diff
@@ -9,6 +9,20 @@
...
+#
+#   # app/controllers/articles_controller.rb
+#   class ApplicationController < ActionController::Base
+#     layout -> { "application" if turbo_frame_request? }
+#   end
+#
+# Then, you have to create a layout with the corresponding name. It can be handy for flash messages, for example:
+#
+#   # app/views/layouts/application.turbo_stream.erb
+#   <%= turbo_stream.update('flash', partial: 'layouts/flash') %>
```
seanpdoyle, line 23: Thank you for proposing this documentation!

The PR's title and the preceding code sample mention Frames, while the layout code sample mentions the `turbo_stream` helper and the template's file extension mentions `turbo_stream.erb`.

As I understand them, Streams and Frames are unrelated (but complementary) tools.

I think the first paragraph could be related to https://github.com/hotwired/turbo-rails/pull/232#issuecomment-1133901506, and feels appropriate in this file.

Is there another, more appropriate area of the codebase we could outline the Streams layout pattern?

alexandreruban: @seanpdoyle, thank you very much for your feedback! 

You are right, *turbo frame requests* and *turbo stream requests* are two different things. In fact, in most cases, we don't want the layout for *turbo frame requests* that are not *turbo stream requests*, as we are only going to extract and replace one single turbo frame. 

However, we want to have this `application.turbo_stream.erb` layout in the case of a *turbo stream request* for flash messages. Maybe we could add a method to identify *turbo stream requests* like this: 

```rb
def turbo_stream_request?
  request.accept.starts_with?(Mime[:turbo_stream])
end
```

Then we could only remove the layout for *turbo frame requests* that are not *turbo stream requests*:

```rb
layout -> { (turbo_frame_request? && !turbo_stream_request?) ? false : "application" }
```

Does it make sense? I just added a new commit so we can discuss this, let me know what you think! If we agree on this, maybe we can then improve the documentation even further.

seanpdoyle: Thank you for clarifying @alexandreruban.

I was under the impression that [a `app/views/layouts/application.turbo_stream.erb` layout is already supported] (https://github.com/hotwired/turbo-rails/issues/25#issuecomment-815774859). Has that changed?

alexandreruban: @seanpdoyle sorry for the late reply. 

The `application.turbo_stream.erb` layout is supported when we override the `layout -> { false if turbo_frame_request? }` of the `turbo-rails` gem with `layout "application"` in our `ApplicationController` for example.

However, this has an unintended side effect. Now, when making GET requests by clicking links inside a Turbo Frame, the view is rendered with the `application.html.erb` layout. We don't need any layout in that case because Turbo is going to replace only the content of a specific Frame.

In fact, what we often want is to add an `application.turbo_stream.erb` layout for Turbo Stream requests (for flash messages most of the time) but not when we make Turbo Frame requests that are not Turbo Stream requests.

I created a [reproduction repository](https://github.com/alexandreruban/layout-reproduction) to clearly show what I mean. The description each commit is very precise and will show you precisely the issue.

seanpdoyle: @alexandreruban thank you for sharing that reproduction repository. 
I think I understand the issue now. I really appreciate the granularity of the commits, it helped me tell the difference between the code generated by Rails and the code that was pertinent to the issue.

app/controllers/turbo/requests.rb
```diff
@@ -9,7 +9,21 @@
 # Turbo Frames knows how to fish out the relevant frame regardless.
 #
 # This module is automatically included in <tt>ActionController::Base</tt>.
-module Turbo::Frames::FrameRequest
+#
+# You can override this behavior in <tt>ApplicationController</tt> if you want to have a layout for turbo stream requests:
+#
+#   # app/controllers/articles_controller.rb
+#   class ApplicationController < ActionController::Base
+#     layout -> { (turbo_frame_request? && !turbo_stream_request?) ? false : "application" }
```
seanpdoyle, line 17: In most cases, Turbo will submit form requests with the Turbo Stream `Accept:` header present, so `turbo_stream_request?` would evaluate to `true`, and any response would be rendered with the `"application"` layout.

Unfortunately, whether or not a response uses a `.html.erb` or `.turbo_stream.erb` template hinges on the **response**, and not the **request**.

For example, consider a `POST` request to `/messages` with an `Accept:` header containing the Turbo Stream MIME type, and a `Turbo-Frame: my-frame` header. The presence of both headers means that:

* `turbo_frame_request?` evaluates to `true`
* `turbo_stream_request?` evaluates to `true`
* `!turbo_stream_request?` evaluates to `false`
* `turbo_frame_request? && !turbo_stream_request?` evaluates to `false`

That means that the `layout` expression would return `"application"`.

Consider a hypothetical `messages#create` controller action that supports both HTML and Turbo Stream responses:

```ruby
def create
  @message = Message.new(messages_params)

  if @message.save
    respond_to do |format|
      format.html { redirect_to @message, status: :see_other }
      format.turbo_stream
    end
  else
    render :new, status: :unprocessable_entity
  end
end
```

In the case of the `format.turbo_stream` block, calls to `turbo_stream_request?` would return `true`, and the documentation's guidance makes sense: render the `messages/create.turbo_stream.erb` template into the `app/views/layouts/application.turbo_stream.erb` layout.

However, since the `POST` request's `Accept:` header includes the Turbo Stream MIME type, calls to `turbo_stream_request?` in the `format.html` branch would also evaluate to `true`. If the subsequent request that follows the redirect is made with the same header value (`Turbo-Frame` and `Accept`, for example) as the original `POST` request, then calls to `turbo_stream_request?` and `turbo_frame_request?` within that request would also evaluate to `true`. This means that no matter what, responses are rendered into an `"application"` layout.

Now consider a `messages#create` controller action that only responds with HTML:

```diff
 def create
   @message = Message.new(messages_params)
 
   if @message.save
-    respond_to do |format|
-      format.html { redirect_to @message, status: :see_other }
-      format.turbo_stream
-    end
+    redirect_to @message
   else
     render :new, status: :unprocessable_entity
   end
 end
```

The absence of a [respond_to](https://edgeapi.rubyonrails.org/classes/ActionController/MimeResponds.html#method-i-respond_to) call combined with the `redirect_to @message` means that the response is going to have an HTML MIME type, not a Turbo Stream MIME type. In that scenario, the `layout` block's inclusion of `!turbo_stream_request?` would still cause the response body to render with the `application` layout.

Personally, I'm in favor of [always rendering responses into a layout](https://github.com/hotwired/turbo-rails/pull/232), but that would require a broader discussion than this one.

Am I misunderstanding something about this process, or something about your particular use case that makes this scenario a non-issue?

alexandreruban: Thanks again for your response. I agree with everything you said above, and you are right for the `redirect_to @message`, I forgot to consider this case as I rarely use it.

The use case was consider a `MessagesController#new` action: 

```rb
def new
  @message = Message.new
end
```

If the link to the new message page is in a Turbo Frame, we would have the following headers: 

```txt
Accept: text/html, application/xhtml+xml
Turbo-Frame: new_article
```

This means that `layout -> { (turbo_frame_request? && !turbo_stream_request?) ? false : "application" }` would evaluate to `false` and so in that specific case, we wouldn't have a layout.

I will close the PR as it does not work in all cases. Thanks again for your review!

dhh: I think the answer is that we need a way at a top-level to set whether you want to use layouts for turbo frames (rendering them in layouts/application.turbo_frame.html) and then document how to overwrite that on a per-action basis with `format.turbo_frame { render layout: "something_else" }`.

PR #355, opened 2022-07-01
Add aria-live to assist screen readers for turbo frame

georgedrummond:
For accessibility, there is an `aria-live` attribute that tells screen readers that a region of the page is "live" and how to announce updates for screen readers.

Right now it is up to the developer to manually enter the `aria-live` attribute when using the frames helper tag.

I propose we introduce `aria-live="polite"` as a sensible default and make accessibility a priority.

* [MDN docs on aria-live](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)

### DIFF
--- a/app/helpers/turbo/frames_helper.rb
+++ b/app/helpers/turbo/frames_helper.rb
@@ -5,40 +5,45 @@ module Turbo::FramesHelper
   # === Examples
   #
   #   <%= turbo_frame_tag "tray", src: tray_path(tray) %>
-  #   # => <turbo-frame id="tray" src="http://example.com/trays/1"></turbo-frame>
+  #   # => <turbo-frame id="tray" src="http://example.com/trays/1" aria-live="polite"></turbo-frame>
   #
   #   <%= turbo_frame_tag tray, src: tray_path(tray) %>
-  #   # => <turbo-frame id="tray_1" src="http://example.com/trays/1"></turbo-frame>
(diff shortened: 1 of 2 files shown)


app/helpers/turbo/frames_helper.rb
```diff
@@ -5,40 +5,45 @@ module Turbo::FramesHelper
...
   #
   #   <%= turbo_frame_tag(Article.find(1), Comment.new) %>
-  #   # => <turbo-frame id="article_1_new_comment"></turbo-frame>
+  #   # => <turbo-frame id="article_1_new_comment" aria-live="polite"></turbo-frame>
   def turbo_frame_tag(*ids, src: nil, target: nil, **attributes, &block)
     id = ids.map { |id| id.respond_to?(:to_key) ? ActionView::RecordIdentifier.dom_id(id) : id }.join("_")
     src = url_for(src) if src.present?
 
-    tag.turbo_frame(**attributes.merge(id: id, src: src, target: target).compact, &block)
+    attrs = attributes.merge(id: id, src: src, target: target)
```
georgedrummond, line 42: This looks a little messy - I'm not sure if there is a cleaner way to deduplicate `aria-live` being passed in through hash or a string.

dhh:
I think this is too far into app-specific territory on how you want to deal with accessibility. I could well imagine a bunch of turbo frames you're just using for the purpose on initial lazy loading, but that you don't want calling out all at once when they load together with the first content (and they're not individually interactive).

PR #356, opened 2022-07-04
Add the `turbo_stream.empty` method

alexandreruban:
We often need to update a part of the DOM with an empty content. It is the case for example when we submit a form with valid attributes, then we want to empty the content of the form and add the created resource to a list on the page.

This PR adds syntactic sugar so that the two following expressions are equivalent, but the second one has a nicer API:

```rb
turbo_stream.update Article.new, ""
turbo_stream.empty Article.new
```

### DIFF
--- a/app/models/turbo/streams/tag_builder.rb
+++ b/app/models/turbo/streams/tag_builder.rb
@@ -154,6 +154,21 @@ class Turbo::Streams::TagBuilder
     action_all :update, targets, content, **rendering, &block
   end
 
+  # Empty the <tt>target</tt> in the dom by removing its content. Example:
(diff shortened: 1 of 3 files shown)


dhh:
I think this is a good instinct. It would normally be the kind of syntactic sugar I'd enjoy. But I think the counterweighting factor here is that the small set of turbo stream actions has value in and off themselves. It reminds me of how we for a while had a huge proliferation of asserts in Rails. All sorts of assert_member_of or whatever. It turned out that the value of such sugar was offset by the increase in API scope. This reminds me of that, so on those grounds, I'm going to pass.

PR #358, opened 2022-07-11
Add support for broadcasting rendering objects

jklina:
Rails supports [rendering of objects](https://edgeguides.rubyonrails.org/layouts_and_rendering.html#rendering-objects) such as [ViewComponents](https://github.com/github/view_component), but this support doesn't seem to be available in Turbo. This is just a stake in the sand for exploring the option to render an `object` in addition to `html` and `partial`.

This should allow for the following API call:

```ruby
message.broadcast_update_to(:messages, target: "message-count",
  object: MessagesCountComponent.new(count: Message.count))
```

where `MessagesCountComponent` is a supported Rails rendering object such as a ViewComponent.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,4 +27,5 @@ group :test do
   gem 'selenium-webdriver'
   gem 'webdrivers'
   gem 'sqlite3'
+  gem 'view_component'
 end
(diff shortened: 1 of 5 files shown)


dhh:
Hmm, rather than adding this custom fork here, I think we're actually better off making it possible in Rails to do `render object: X` in addition to `render X`. Would match with the other long forms like `render partial: X`. Let's pursue that there. Feel free to reference this comment here as approval to pursue.

PR #360, opened 2022-07-14
Enhance provided Drive helpers to allow controlling `cache` and `visit` meta tags

agrobbin:
Being able to control these `<meta>` tags with helper methods has proved really valuable in a project of ours, and seems like it would be useful for others!

I've deprecated `turbo_exempts_page_from_cache` in favor of `turbo_cache_control :no_cache` as well, since they do the same thing.

### DIFF
--- a/app/helpers/turbo/drive_helper.rb
+++ b/app/helpers/turbo/drive_helper.rb
@@ -1,6 +1,5 @@
 module Turbo::DriveHelper
-  # Pages that are more likely than not to be a cache miss can skip turbo cache to avoid visual jitter.
-  # Note: This requires a +yield :head+ provision in the application layout.
+  # Note: These helpers require a +yield :head+ provision in the layout.
   #
   # ==== Example
   #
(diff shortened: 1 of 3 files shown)


dhh:
There are not enough permutations to warrant the generalized approach, imo. I'd rather just see us add `turbo_exempts_page_from_preview` and `turbo_forced_reload` ala the existing `turbo_exempts_page_from_cache`.

agrobbin:
I don't have a really strong opinion here. My 2 reasons for doing it this way were to align the implementation and interface w/ the actual `<meta>` tags that are produced, and to make it really clear to devs that the `no-cache` and `no-preview` scenarios are exclusive of one another. If I saw this in a template:

```erb
<% turbo_exempts_page_from_preview -%>
<% turbo_exempts_page_from_cache -%>
```

I would not immediately think there's a problem, whereas if I saw:

```erb
<% turbo_cache_control :no_preview -%>
<% turbo_cache_control :no_cache -%>
```

I would think "huh, that seems suspect, I should check what happens if I do both".

That said, if you feel strongly @dhh, I'm happy to define the 3 explicitly!

dhh:
I think we should just tackle just with a documentation point. Just finding the weight of the implementation a bit disproportionate to what the outcome is (a meta tag).

app/helpers/turbo/drive_helper.rb
```diff
@@ -10,7 +9,21 @@ module Turbo::DriveHelper
...
+
+  # Specify that a cached version of the page should not be shown as a preview during an application visit.
+  # Cannot be used along with +turbo_exempts_page_from_cache+.
+  def turbo_exempts_page_from_preview
+    provide :head, tag.meta(name: 'turbo-cache-control', content: 'no-preview')
+  end
+
+  # Force the page, when loaded by Turbo, to be cause a full page reload.
+  def turbo_page_requires_reload
+    provide :head, tag.meta(name: 'turbo-visit-control', content: 'reload')
```
dhh, line 27: Should use " instead of ' everywhere for strings.

agrobbin:
👍🏻 just pushed an update!

agrobbin:
🥳 thanks for the quick review and merge, @dhh!

PR #371, opened 2022-08-12
Add Empty Template Tags To The Remove Action

MatthewKennedy:
While using the beta version of [hot-wired/turbo](https://github.com/hotwired/turbo), it was noted that the remove action now requires a child element <template> tag. 

If this is the intended behaviour, this PR updates the turbo_stream tag helper to add in the empty <template>  tags.


marcoroth:
Hey @MatthewKennedy, thanks for opening this PR!

I don't think this is still necessary, since this was fixed via https://github.com/hotwired/turbo/pull/665

PR #372, opened 2022-08-19
Fix compatibility shim for mobile apps

okcomputer93:
This fixes an issue in the suggested compatibility shim for mobile apps,
when the adapter is handled by turbolinks-ios or turbolinks-android.
Because the reference to `adapter.visitStarted` is  stored as a
variable, when it's called again the reference to `this` is not pointing
to the adapter's `this` but to `window`.

Co-authored-by: Juan Zenón <juanzv@hey.com>

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -135,7 +135,7 @@ window.Turbolinks = {
 
         // Old mobile adapters use visit.location.absoluteURL, which is not available
         // because Turbo dropped the Location class in favor of the DOM URL API
-        const adapterVisitStarted = adapter.visitStarted
+        const adapterVisitStarted = adapter.visitStarted.bind(adapter)
         adapter.visitStarted = function(visit) {
           Object.defineProperties(visit.location, {
             absoluteURL: {


UPGRADING.md
```diff
@@ -135,7 +135,7 @@ window.Turbolinks = {
 
         // Old mobile adapters use visit.location.absoluteURL, which is not available
         // because Turbo dropped the Location class in favor of the DOM URL API
-        const adapterVisitStarted = adapter.visitStarted
+        const adapterVisitStarted = adapter.visitStarted.bind(adapter)
```
okcomputer93, line 138: We're having this issue while trying to migrate to Turbo from our apps in `turbolinks-ios`,when the `visitStarted` method in the adapter is called, `this` is a reference to `window` instead of the adapter itself:
https://github.com/turbolinks/turbolinks-ios/blob/1c891b9ffbb89666a05b7c5b02a8fdd700e92323/Turbolinks/WebView.js#L60-L63

dhh:
cc @jayohms

PR #373, opened 2022-08-20
Allow additional attributes on the `turbo_stream_action_tag` helper

marcoroth:
Turbo now supports Custom Stream Actions via https://github.com/hotwired/turbo/pull/479. In order to be able to properly use Custom Stream Actions in your application you also need to have a new server-side helper to generate the corresponding `<turbo-stream>` element.

Imagine a Custom Stream Action `console_log` which prints out a message via JavaScript's `console.log`. The action takes two arguments: the first one is the message itself and the second optional argument is the log-level.

```js
import { StreamActions } from '@hotwired/turbo'

StreamActions.console_log = function () {
  const level = this.getAttribute('level') || 'log'
  const message = this.getAttribute('message')

  console[level](message)
}
```

This Pull Requests allows to write a custom helper, like so:

```ruby
# app/helpers/custom_stream_helper.rb

module CustomStreamHelper
  def console_log(message, level = :log)
    turbo_stream_action_tag :console_log, target: "body", message: message, level: level
  end
end

Turbo::Streams::TagBuilder.prepend(CustomStreamHelper)
```

This now allows you to call `console_log` on `turbo_stream` throughout the application with:
```ruby
turbo_stream.console_log "Hello World"
```

Which produces the following payload:
```html
<turbo-stream message="Hello World" level="log" action="console_log" target="body"><template></template></turbo-stream>
```

Previously this wasn't possible because the `turbo_stream_action_tag` method wouldn't take any additional arguments.

### DIFF
--- a/app/helpers/turbo/streams/action_helper.rb
+++ b/app/helpers/turbo/streams/action_helper.rb
@@ -15,9 +15,9 @@ module Turbo::Streams::ActionHelper
     template = action.to_sym == :remove ? "" : tag.template(template.to_s.html_safe)
 
     if target = convert_to_turbo_stream_dom_id(target)
-      tag.turbo_stream(template, **attributes.merge(action: action, target: target).compact)
+      tag.turbo_stream(template, **attributes.merge(action: action, target: target))
     elsif targets = convert_to_turbo_stream_dom_id(targets, include_selector: true)
-      tag.turbo_stream(template, **attributes.merge(action: action, targets: targets).compact)
+      tag.turbo_stream(template, **attributes.merge(action: action, targets: targets))
     else
       raise ArgumentError, "target or targets must be supplied"
     end
--- a/app/helpers/turbo/streams/action_helper.rb
+++ b/app/helpers/turbo/streams/action_helper.rb
@@ -1,4 +1,6 @@
 module Turbo::Streams::ActionHelper
+  include ActionView::Helpers::TagHelper
+
   # Creates a `turbo-stream` tag according to the passed parameters. Examples:
   #
   #   turbo_stream_action_tag "remove", target: "message_1"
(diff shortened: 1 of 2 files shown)


app/helpers/turbo/streams/action_helper.rb
```diff
@@ -9,13 +11,13 @@ module Turbo::Streams::ActionHelper
   #
   #   turbo_stream_action_tag "replace", targets: "message_1", template: %(<div id="message_1">Hello!</div>)
   #   # => <turbo-stream action="replace" targets="message_1"><template><div id="message_1">Hello!</div></template></turbo-stream>
-  def turbo_stream_action_tag(action, target: nil, targets: nil, template: nil)
-    template = action.to_sym == :remove ? "" : "<template>#{template}</template>"
+  def turbo_stream_action_tag(action, target: nil, targets: nil, template: nil, **attributes)
+    template = action.to_sym == :remove ? "" : tag.template(template.to_s.html_safe)
```
seanpdoyle, line 15: Could `tag.template(template.to_s.html_safe)` be replaced by `tag.template(template)`? 

I'm able to render this without issue:

```erb
<%= tag.template(nil) %>
```

marcoroth: That would be finr for the nil/empty case, but if there is a template it's going to be escaped.

app/helpers/turbo/streams/action_helper.rb
```diff
@@ -9,13 +11,13 @@ module Turbo::Streams::ActionHelper
...
 
     if target = convert_to_turbo_stream_dom_id(target)
-      %(<turbo-stream action="#{action}" target="#{target}">#{template}</turbo-stream>).html_safe
+      tag.turbo_stream(template, **attributes.merge(action: action, target: target).compact)
     elsif targets = convert_to_turbo_stream_dom_id(targets, include_selector: true)
-      %(<turbo-stream action="#{action}" targets="#{targets}">#{template}</turbo-stream>).html_safe
+      tag.turbo_stream(template, **attributes.merge(action: action, targets: targets).compact)
```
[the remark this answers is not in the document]
marcoroth: Good call. I think we might be able to remove it!

app/helpers/turbo/streams/action_helper.rb
seanpdoyle: Ah, I understand now. Thank you for clarifying.

Since we're expanding the interface of this helper, how do you feel about also accepting the `template` value as a block, like most other Action View tag helpers?

```ruby
template = block.present? ? capture(&block) : template.to_s.html_safe
# ...
```

marcoroth: I was also thinking about that. I first looked at how to extend the [`TagBuilder#action` method](https://github.com/hotwired/turbo-rails/blob/main/app/models/turbo/streams/tag_builder.rb#L214-L218) which would already support that, but since it's already accepting the `rendering` arguments for rendering the template I figured it would be simpler to extend the `turbo_stream_action_tag` with just the additional attributes.

Or do you see a non-hacky way to adjust the `action` method to take additional attributes?

I don't think this is a good idea:
```diff
  def action(name, target, content = nil, allow_inferred_rendering: true, **rendering, &block)
    template = render_template(target, content, allow_inferred_rendering: allow_inferred_rendering, **rendering, &block)
   
-    turbo_stream_action_tag name, target: target, template: template
+    turbo_stream_action_tag name, target: target, template: template, **rendering
  end
```

PR #374, opened 2022-08-22
Allow Turbo Stream actions to be provided as keyword arguments

marcoroth:
This pull request allows methods within the `Turbo::Streams::TagBuilder` class to be used with keywords arguments. This also allows users to build helpers for custom action helpers when the `target` is not required, as seen in #373.

It also adds tests cases for all seven default actions and the `action` and `action_all` helper methods to make sure we support all possible variants.

The amount of times I've seen people struggle with the `target` and `targets` attribute is super high. I'm also running into it from time to time. I always found it confusing to have a regular action helper for `target` and a `*_all` method for `targets`.

What this pull request proposes is to use keywords argument to make it more readable and clear which attribute is being used.

```ruby
turbo_stream.append(target: "messages", content: "Append")
turbo_stream.append(targets: "messages", content: "Append")

# => <turbo-stream action="append" target="messages"><template>Append</template></turbo-stream>
# => <turbo-stream action="append" targets="messages"><template>Append</template></turbo-stream>
```

 This change will not break backwards compatibility, as it accounts for the arguments to be passed as positional or keyword arguments. So this is also still valid:

```ruby
turbo_stream.append("messages", "Append")
turbo_stream.append_all("messages", "Append")

# => <turbo-stream action="append" target="messages"><template>Append</template></turbo-stream>
# => <turbo-stream action="append" targets="messages"><template>Append</template></turbo-stream>
```

The other problem the keywords argument solve is that you don't need to remember which argument is which and in which order you need to pass them. With keywords arguments you can pass them in any order. This might not be a big issue for the default actions, but is really beneficial for custom actions.

This also helps when building your own helpers for custom action without having to [reinvent the wheel](https://github.com/marcoroth/turbo_power-rails/blob/68d58315ed4843c86418a29de9b12a06961a0006/lib/turbo_power/stream_helper.rb#L10-L18). Previously you had to do something like:
```ruby
module TurboStreamActionsHelper
  def morph(target, content, children_only = true)
    turbo_stream_action_tag(
      :morph,
      target: target,
      template: content,
      children_only: children_only
    )
  end

  def morph_all(targets, content, children_only = true)
    turbo_stream_action_tag(
      :morph,
      targets: targets,
      template: content,
      children_only: children_only
    )
  end
end
```

With this pull request you can now make use of the `action` method and also get the `target` vs `targets` for free, without having to declare two methods in your helper for the custom action. The other benefit you get from this is that you can now pass in a block for the rendering or use the rendering mechanism which are built-into the default actions. So one can now implement a helper for a custom action like this:

```ruby
module TurboStreamActionsHelper
  def morph(target: nil, targets: nil, content: nil, **kwargs, &block)
    action :morph, target: target, targets: targets, content: content, attributes: kwargs, &block)
  end
end
```

Which allows this action to be used in all variations:
```html+erb
<%= turbo_stream.morph(target: "posts", content: "Posts") %>
<%= turbo_stream.morph(targets: ".post", content: "Post") %>

<%= turbo_stream.morph(target: dom_id(@post), partial: "posts/post", locals: { post: @post } ) %>

<%= turbo_stream.morph(target: dom_id(@post)) do %>
  <h1><%= @post.title %></h1>
<% end %>
```

To achieve this previously it required quite a sophisticated helper method. This pull request really simplifies this a lot!

Related: #375

### DIFF
--- a/app/models/turbo/streams/tag_builder.rb
+++ b/app/models/turbo/streams/tag_builder.rb
@@ -37,7 +37,7 @@ class Turbo::Streams::TagBuilder
   #   <%= turbo_stream.remove "clearance_5" %>
   #   <%= turbo_stream.remove clearance %>
   def remove(target)
-    action :remove, target, allow_inferred_rendering: false
+    action :remove, target: target, allow_inferred_rendering: false
   end
 
   # Removes the <tt>targets</tt> from the dom. The targets can either be a CSS selector string or an object that responds to
@@ -47,7 +47,7 @@ class Turbo::Streams::TagBuilder
   #   <%= turbo_stream.remove_all ".clearance_item" %>
   #   <%= turbo_stream.remove_all clearance %>
   def remove_all(targets)
-    action_all :remove, targets, allow_inferred_rendering: false
+    action_all :remove, targets: targets, allow_inferred_rendering: false
   end
 
   # Replace the <tt>target</tt> in the dom with either the <tt>content</tt> passed in, a rendering result determined
@@ -60,7 +60,7 @@ class Turbo::Streams::TagBuilder
   #     <div id='clearance_5'>Replace the dom target identified by clearance_5</div>
   #   <% end %>
   def replace(target, content = nil, **rendering, &block)
-    action :replace, target, content, **rendering, &block
+    action :replace, target: target, content: content, **rendering, &block
   end
 
   # Replace the <tt>targets</tt> in the dom with either the <tt>content</tt> passed in, a rendering result determined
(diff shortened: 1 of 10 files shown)


dhh:
Wouldn't this be a breaking change?

marcoroth:
I guess that depends if we treated `action` and `action_all` as part of the public API or not. Technically there was no reason to call those methods directly. Just now with 7.2 and the introduction of Custom Actions people now have a reason to call those methods directly.

marcoroth:
I made this change backwards compatible by allowing the arguments to be either passed as positional arguments or keyword arguments.

dhh:
Hmm, I appreciate where this is coming from, but I don't actually like ending up with two simultaneous API styles for these invocations. Possible that it would have been preferential to go with kwargs originally, but given that we didn't, I don't think there's enough value to going there now, ending up with forked styles.

PR #375, opened 2022-08-22
Make `target` and `targets` attributes optional for Stream actions

marcoroth:
With the introduction of Custom Stream Actions we can now create Stream Actions that not necessarily require  a `target` attribute to be present.

~~In order to keep the behaviour we need to raise an `ArgumentError` if no `target` was passed to a default stream action helper.~~

~~This Pull Request adds a constant with the default actions which require a `target` or `targets` attribute to be present. For any other action it will return a regular `<turbo-stream>` tag without the `target` attribute.~~

Edit: This Pull Request removes the check and thus makes the `target` and `targets` attributes optional for stream actions.

Related: #373 and #374

### DIFF
--- a/app/helpers/turbo/streams/action_helper.rb
+++ b/app/helpers/turbo/streams/action_helper.rb
@@ -19,7 +19,7 @@ module Turbo::Streams::ActionHelper
     elsif targets = convert_to_turbo_stream_dom_id(targets, include_selector: true)
       tag.turbo_stream(template, **attributes.merge(action: action, targets: targets))
     else
-      raise ArgumentError, "target or targets must be supplied"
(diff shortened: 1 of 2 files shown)


dhh:
Don't actually think we need this degree of error checking. We're just generating an HTML tag here. Fine to simply allow it not to have a target, and thus be invalid, if the action requires a target.

marcoroth:
I just added the check so that we can remain the behavior between versions. Because applications could technically rely on this error to be thrown, but I would be fine to remove the check entirely.

dhh:
I'd say just remove the check. Don't see how an app will rely on this. It's a failure state.

PR #376, opened 2022-08-24
Call `to_turbo_stream` on the object passed to the Turbo Stream renderer

marcoroth:
Currently the `turbo_stream` format in the controller action just accepts HTML in form of a string.

This Pull Requests looks if the passed object to the Turbo Stream renderer responds to `.to_turbo_stream` and if it does calls the method on it.

This allows developers to pass in any object as long as it responds to the right method, which allows for some more advanced  Turbo Stream use-cases to build out the Turbo Stream payload.

### DIFF
--- a/lib/turbo/engine.rb
+++ b/lib/turbo/engine.rb
@@ -55,11 +55,8 @@ module Turbo
       ActiveSupport.on_load(:action_controller) do
         ActionController::Renderers.add :turbo_stream do |turbo_streams_html, options|
           self.content_type = Mime[:turbo_stream] if media_type.nil?
-          if turbo_streams_html.respond_to?(:to_turbo_stream)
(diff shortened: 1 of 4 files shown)


marcoroth:
Of course, this is a rather simple example. I mainly added this so that you could build a class which builds out a Turbo Streams response with a more complex setup than just calling `.new` on it.

For example, if I would want to build an app-specific PORO which abstracts away some business logic to generate a set of related Turbo Stream actions it would be nice if it automatically tries to  call `to_turbo_stream` on the passed object.

dhh:
Appreciate the sentiment, and there is some precedence around stuff like #to_json, but I think in this case, because turbo stream is not a generically useful format, like json, I don't think the overall balance is worth adding the convention.

PR #384, opened 2022-09-14
Don't assume fetchOptions has a `body`

kevinmcconnell:
The event handler in `encodeMethodIntoRequestBody` was failing when forms were submitted using `method="GET"`, because it assumed the presence of `fetchOptions.body`, and such forms don't have a body.

Checking for the presence of `body` before accessing it avoids the crash, and the resulting console noise.


dhh:
Need to run `yarn build` and checkin the assets with changes 👍

kevinmcconnell:
Just realised that too :) Thanks!

Clearly I've become too used to CI doing that sort of thing for me.

PR #392, opened 2022-10-08
Convert JavaScript To TypeScript

kzkn:
For fix #18

PR #403, opened 2022-11-18
Add Type Safety Guards to `turbo/fetch_requests`

seanpdoyle:
Without this change, GET `<form>` submissions without a Fetch Options `{
body: }` raise the following:

```
Uncaught TypeError: s.body is null
    <anonymous> fetch_requests.js:10
    w turbo.es2017-esm.js:351
    requestStarted turbo.es2017-esm.js:770
    perform turbo.es2017-esm.js:520
    start turbo.es2017-esm.js:744
    formSubmitted turbo.es2017-esm.js:3369
```

Through the exercise of attempting to [port the `turbo/fetch_requests` module to
TypeScript](https://github.com/hotwired/turbo-rails/pull/392#issuecomment-1320261244), we've identified some potential edge cases in the algorithm that determines a request's HTTP method.

Even if the migration to TypeScript doesn't come to fruition for some time, those edge cases should be addressed sooner rather than later.

This commit adds more "type safety" motivated conditionals and guards to make sure that values are present before overriding them.


seanpdoyle:
cc: @dhh

kzkn:
@seanpdoyle 
Thanks for the comment.
I was having trouble sorting this part out. Your code looks better than mine 😀

#403 had already been merged into the main branch, so I will try to rewrite it using that as a reference.

PR #408, opened 2022-12-09
Support `targets:` in `Broadcastable` methods

ghiculescu:
https://github.com/hotwired/turbo-rails/pull/210 added the ability to set `targets` on a stream tag. But that doesn't work nicely with the `Broadcastable` helper methods. Currently you have to do this to target some `targets`:

```ruby
after_update_commit -> { broadcast_update_to self, target: nil, targets: ".class_name" }
```

With this PR, you don't need to provide `target: nil` anymore.

Notes:

- Should this be covered in the comments? Any preference on where?
- I didn't update `broadcasts_to` and `broadcasts`, because they have special handling for procs only on creates. I was struggling to get that logic right. I'm also not sure *why* that special handling only exists on creates?

### DIFF
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -96,13 +96,13 @@ module Turbo::Broadcastable
   #
   #   # Sends <turbo-stream action="remove" target="clearance_5"></turbo-stream> to the stream named "identity:2:clearances"
   #   clearance.broadcast_remove_to examiner.identity, :clearances
-  def broadcast_remove_to(*streamables, target: self)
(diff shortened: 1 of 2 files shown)


ghiculescu:
Those red tests feel flakey to me :/

SebastianSzturo:
Anything I can help to fix this?

JuannFerrari:
This would be nice to have

ghiculescu:
I'm not going to have time to fix those merge conflicts anytime soon. If someone wants to make a new PR I'm fine with that.

brunoprietog:
Done in #511

PR #411, opened 2022-12-13
Allow passing extra params to native redirect URL

kevinmcconnell:
When using the various redirect helpers, we allow passing the `:notice` option to the native redirection URL as a query param. However in some cases, apps may need to pass additional query parameters as well.
    
This adds an optional `:native_params` option, which can be a hash of query params to add to the URL when redirecting on native. This option is ignored on other platforms.
    
We also drop the special-case handling of `:native_notice` here, as that can now be more consistently handled inside `:native_params`.


jayohms (approved):
Lovely, thank you 👏

PR #413, opened 2022-12-24
Parallelize test suite

seanpdoyle:
In the past, the test suite has been flaky, and didn't seem to truncate tables from previous runs.

The root cause of that might be related to [rails/rails#46820][]. In the meantime, while that is resolved, this commit adds a line to [parallelize][] to the test suite, and some corresponding `.sqlite3` ignore values to the `.gitignore`.

[rails/rails#46820]: https://github.com/rails/rails/issues/46820
[parallelize]: https://edgeapi.rubyonrails.org/classes/ActiveSupport/TestCase.html#method-c-parallelize

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.gem
 /node_modules
 Gemfile.lock
+*.sqlite3*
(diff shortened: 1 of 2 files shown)


seanpdoyle:
Hm, this passes consistently when executed locally.

PR #417, opened 2023-01-06
use helper_method for turbo_frame_request_id

igorkasyanchuk:
I want to remove redundant code from my controller because I need to know turbo_frame_request_id inside the view


dhh:
Can you elaborate on the controller code you were able to simplify with this?

igorkasyanchuk:
it basically removes the variable `@frame_id = turbo_frame_request_id` from the controller. 

In case I need to render partial from different actions, or views I don't need to initialize it in the controller. So we have cleaner code and actually have to write less code.

PR #419, opened 2023-01-26
Expose `Turbo::Native::Navigation#turbo_native_app?` helper method

seanpdoyle:
First, promote `Turbo::Native::Navigation` to be an [ActiveSupport::Concern][].

Next, expose `#turbo_native_app?` to be a view helper through a call to [helper_method][] from within the [included][] block.

[ActiveSupport::Concern]: https://edgeapi.rubyonrails.org/classes/ActiveSupport/Concern.html
[helper_method]: https://edgeapi.rubyonrails.org/classes/AbstractController/Helpers/ClassMethods.html#method-i-helper_method
[included]: https://edgeapi.rubyonrails.org/classes/ActiveSupport/Concern.html#method-i-included

### DIFF
--- a/test/dummy/app/views/layouts/application.html.erb
+++ b/test/dummy/app/views/layouts/application.html.erb
@@ -9,7 +9,7 @@
     <%= javascript_importmap_tags %>
   </head>
 
-  <body>
+  <body class="<%= "turbo-native" if turbo_native_app? %>">
     <%= yield %>
   </body>
 </html>
--- a/app/controllers/turbo/native/navigation.rb
+++ b/app/controllers/turbo/native/navigation.rb
@@ -4,6 +4,12 @@
 #
 # turbo-android handles these actions automatically. You are required to implement the handling on your own for turbo-ios.
 module Turbo::Native::Navigation
+  extend ActiveSupport::Concern
+
+  included do
+    helper_method :turbo_native_app?
+  end
+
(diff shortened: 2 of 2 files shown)


test/dummy/app/views/layouts/application.html.erb
```diff
@@ -9,7 +9,7 @@
     <%= javascript_importmap_tags %>
   </head>
 
-  <body>
+  <body class="<%= "turbo-native" if turbo_native_app? %>">
```
dhh, line 12: Did you mean to add a test to back this up or is this just a smoke test?

seanpdoyle: My plan was to add a test, but then I chased the CI failures across the other PRs. There will be a test to cover this behavior if we deem it viable.

dhh:
What was the use case that triggered the need for this method being public in your app?

seanpdoyle:
The motivation for this feature is rooted in conditionally applying CSS based on whether or not the request is made from the Turbo-enabled Web application, or a Turbo Native-enabled Mobile application. The project used Tailwind for CSS, but the same pattern might emerge in other CSS philosophies. It involved something like:

```html
<style>
  .native .native\:hidden { display: none; }
</style>

<body class="<%= token_list "native" => turbo_native_app? %>">
  <nav class="native:hidden">...</nav>
</body>
```

PR #421, opened 2023-01-29
Force CI execution to be sequential

seanpdoyle:
In the past, the test suite has been flaky, and didn't seem to truncate tables from previous runs.

To attempt to resolve that issue, add a [parallelize][] configuration line to our `ActiveSupport::TestCase`, then force a single sequential test execution in CI by setting `PARALLEL_WORKERS=1`.

[parallelize]: https://edgeapi.rubyonrails.org/classes/ActiveSupport/TestCase.html#method-c-parallelize

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
 
     env:
       RAILS_VERSION: "${{ matrix.rails-version }}"
+      PARALLEL_WORKERS: 1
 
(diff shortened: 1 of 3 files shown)


dhh:
Are the tests on this PR using those new settings?

seanpdoyle:
I've opened a PR to execute tests with more parallel workers, tests in sequence, tests driven by Cuprite. 

No matter what, when they consistently pass locally, they're flaky in the CI matrix.

PR #425, opened 2023-02-02
Allow passing a `:template` option to Broadcastable methods

davidalejandroaguilar:
# Description

## Summary

This PR allows passing the `:template` option to the `**rendering` options of a `Broadcastable` method, so we can render a whole template instead of a partial.

## Rationale

Currently, we can only use `:partial`, `:html` or pass nothing (in this case `Broadcastable` calls the model's `to_partial_path` method). Meaning we're assuming that we'll always be dealing with partials or custom HTML written inline. If we rather want to render a whole template (e.g. index, show, edit, etc.), we need to resort to using `ApplicationController.render`, which is not ideal.

## Notes

When the `template` option is passed, we'll pass it down with the `layout: false` option, in order to avoid passing any unwanted elements such as `<head>`.

### DIFF
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -27,7 +27,7 @@
 # (which is derived by default from the plural model name of the model, but can be overwritten).
 #
 # You can also choose to render html instead of a partial inside of a broadcast
-# you do this by passing the html: option to any broadcast method that accepts the **rendering argument
+# you do this by passing the `html:` option to any broadcast method that accepts the **rendering argument. Example:
 #
 #   class Message < ApplicationRecord
 #     belongs_to :user
@@ -37,6 +37,20 @@
 #     private
 #       def update_message_count
 #         broadcast_update_to(user, :messages, target: "message-count", html: "<p> #{user.messages.count} </p>")
+#       end
+#   end
+#
+# If you want to render a template instead of a partial, e.g. ('messages/index' or 'messages/show'), you can use the `template:` option.
+# Again, only to any broadcast method that accepts the `**rendering` argument. Example:
+#
+#   class Message < ApplicationRecord
+#     belongs_to :user
+#
+#     after_create_commit :update_message
+#
+#     private
+#       def update_message
+#         broadcast_replace_to(user, :message, target: "message", template: "messages/show", locals: { message: self })
 #       end
 #   end
 #
@@ -335,8 +349,13 @@ module Turbo::Broadcastable
         # Add the current instance into the locals with the element name (which is the un-namespaced name)
         # as the key. This parallels how the ActionView::ObjectRenderer would create a local variable.
         o[:locals] = (o[:locals] || {}).reverse_merge!(model_name.element.to_sym => self)
-        # if the html option is passed in it will skip setting a partial from #to_partial_path
-        unless o.include?(:html)
+
+        if o[:html] || o[:partial]
+          return o
+        elsif o[:template]
+          o[:layout] = false
+        else
+          # if none of these options are passed in, it will set a partial from #to_partial_path
           o[:partial] ||= to_partial_path
         end
       end
--- a/test/streams/broadcastable_test.rb
+++ b/test/streams/broadcastable_test.rb
@@ -36,6 +36,12 @@ class Turbo::BroadcastableTest < ActionCable::Channel::TestCase
     end
   end
 
+  test "broadcasting update to stream now with template option" do
+    assert_broadcast_on "stream", turbo_stream_action_tag("update", target: "message_1", template: render("messages/index", layout: false)) do
+      @message.broadcast_update_to "stream", template: "messages/index"
+    end
+  end
+
   test "broadcasting update now" do
     assert_broadcast_on @message.to_gid_param, turbo_stream_action_tag("update", target: "message_1", template: render(@message)) do
       @message.broadcast_update


test/streams/broadcastable_test.rb
```diff
@@ -36,6 +36,12 @@ class Turbo::BroadcastableTest < ActionCable::Channel::TestCase
     end
   end
 
+  test "broadcasting update to stream now with template option" do
+    assert_broadcast_on "stream", turbo_stream_action_tag("update", target: "message_1", template: render("messages/index", layout: false)) do
+      @message.broadcast_update_to "stream", template: "messages/index"
+    end
+  end
```
davidalejandroaguilar, line 43: I'm wondering if we'd want:

* Only one method tested with the `template` option, like the current commit has.
* Separate tests for each `Broadcastable` method. e.g.

```rb
  test "broadcasting update to stream now with template option" do
    assert_broadcast_on "stream", turbo_stream_action_tag("update", target: "message_1", template: render("messages/index", layout: false)) do
      @message.broadcast_update_to "stream", template: "messages/index"
    end
  end

  test "broadcasting replace to stream now with template option" do
    assert_broadcast_on "stream", turbo_stream_action_tag("replace", target: "message_1", template: render("messages/index", layout: false)) do
      @message.broadcast_replace_to "stream", template: "messages/index"
    end
  end
```

* Separate methods in a dynamic way, e.g.:

```rb 
  [
    [:broadcast_update_to, :update],
    [:broadcast_replace_to, :replace]
  ].each do |method, action|
    test "broadcasting #{method} to stream now with template option" do
      assert_broadcast_on "stream", turbo_stream_action_tag(action, target: "message_1", template: render("messages/index", layout: false)) do
        @message.public_send method, "stream", template: "messages/index"
      end
    end
  end

  [
    [:broadcast_before_to, :before],
    [:broadcast_after_to, :after],
    [:broadcast_append_to, :append],
    [:broadcast_prepend_to, :prepend],
  ].each do |method, action|
    test "broadcasting #{method} to stream now with template option" do
      assert_broadcast_on "stream", turbo_stream_action_tag(action, target: "message_1", template: render("messages/index", layout: false)) do
        @message.public_send method, "stream", template: "messages/index", target: "message_1"
      end
    end
  end

  # ...etc.
```

Reason I only added one test for now is because the current tests are not thoroughly testing all possible rendering options for all methods. e.g. the `partial` option is only tested in the test named `local variables don't get overwritten if they collide with the template name`, so I don't know how to proceed.

app/models/concerns/turbo/broadcastable.rb
```diff
@@ -40,6 +40,20 @@
...
+#   class Message < ApplicationRecord
+#     belongs_to :user
+#
+#     after_create_commit :update_message
+#
+#     private
+#       def update_message
+#         broadcast_replace_to(user, :message, target: "message", template: "messages/show", locals: { message: self })
+#       end
+#   end
```
davidalejandroaguilar, line 55: This might not be the best example but I'm just trying to follow the existing "messages" context of the examples.

test/streams/broadcastable_test.rb
dhh: Prefer the separate tests. Please do open a follow up PR to expand test coverage!

kevinmcconnell:
What would the use case be for broadcasting a view template?

I would normally expect the view templates to be rendered as the response to a controller action. When you want to reuse content from a view in multiple scenarios, extracting that content to a partial is a nice convention. It also means you're passing locals rather than ivars to those partials, whereas the views will typically expect state in ivars.

There's probably a good reason for this that I'm missing though!

davidalejandroaguilar:
That's a valid question @kevinmcconnell! Personally, my use case is this:

I have a turbo-frame modal that works as a stepper/wizard. It has a `src` attribute that, on page load, hits the `show` action for a specific route and displays the contents of step 1. Upon clicking "Continue" from within the modal, the `show` action of step 2's route is hit and the modal is updated with this page. And so on and so forth.

In this case, the `show` actions are rendering the whole `show` template, they're not using partials. So when using `Broadcastable` to update that modal's contents with new information, I'd have to either:

1. Have my `show.html.erb` template only do a `render "show"` that renders a `_show.html.erb` partial with the actual content and have the broadcast method use `partial: "show"`.
2. Use `ApplicationController.render` to render `show.html.erb` and pass it to the broadcast method `html:` option.

Having the option to use `template` would prevent me from wrestling the framework in order to achieve the above.

Now, aside from my use case. I think this change is also valid for 2 reasons:

* Principle of least surprise: If I see that the `partial:` and `html:` options are available and that they're ultimately going through `ApplicationController.render`. I'd think using `template:` would _just work_.
* Ruby giving you _sharp knives_: Having the option to render a template is a powerful tool. But it doesn't mean it is the right one for every job. So as with everything in Ruby, you've got to use your judgement here.

Finally, regarding the use of locals rather than ivars. I personally always use locals/helper methods rather than ivars. Since if an ivar is not defined, Ruby will not complain, and so you might end up with `nil` values where you don't expect them. On the other hand, if you use locals/helper methods, Ruby will complain loudly when you try to use them and they're not defined.

However, since this these options are all going to be passed down to `ApplicationController.render` at the end (on `Turbo::Streams::Broadcasts#render_format`), you can either use `locals:` or `assigns:` to pass either local variables/methods or ivars. Maybe we should use `assigns` too in the docs example?

davidalejandroaguilar:
@dhh  Failing CI tests pass on my local, e.g.

image

I've seen other PRs merged with some failing CI jobs, so not sure what the next steps are.

dhh:
Yeah CI is flaky. Will fix elsewhere.

kevinmcconnell:
@davidalejandroaguilar thanks for the explanation!

I'd still use a partial for that case myself (but not called `_show.html.erb`; probably something more like `_stage.html.erb` or whatever your domain language suggests). It helps communicate that this is a shared template used in multiple ways. But I get what you're saying. Thanks for satisfying my curiosity :)

> Maybe we should use assigns too in the docs example?

I think that's a great idea 👍

PR #427, opened 2023-02-04
Allow broadcasting without rendering and broadcasting with custom attributes

davidalejandroaguilar:
# Description

## Key Enhancements

This PR improves the flexibility of custom Turbo Stream actions by allowing:

* Broadcasting without rendering
  * Some custom actions don't need any rendering at all because they're [custom actions](https://turbo.hotwired.dev/handbook/streams#custom-actions). 
* Enabling passing custom attributes via the `Broadcastable` concern.
  * These custom actions might need them.   

## Motivation

Thanks to the work on https://github.com/hotwired/turbo/pull/479 and https://github.com/hotwired/turbo-rails/pull/373,  custom `turbo_stream` actions can now be created. However, some limitations still exist ([See here for an example of both current limitations](https://github.com/hotwired/turbo-rails/pull/373)):

- Limited flexibility for passing extra attributes
  - Currently, extra attributes can only be passed via `turbo_stream_action_tag` and `turbo_stream`. 
- Inability to use custom actions via models (which use the `Broadcastable` concern)
  - Broadcasting custom actions to any streamable/model is not possible. It can only be done via a `turbo_stream` response on the controller or inside a view response (`.turbo_stream.erb`) for a single user.

## Example

A [custom stream action](https://turbo.hotwired.dev/handbook/streams#custom-actions) that doesn't need a template, only attributes (e.g. a toast):

```js
StreamActions.toast = function () {
  const text = this.getAttribute('text');
  const duration = Number(this.getAttribute('duration') || 3000);

  const toast = Toastify({
    text,
    duration,
  });

  toast.showToast();
};
```

Create a module with custom broadcast methods, just to encapsulate them and include them wherever we want:

```rb
module TurboBroadcastable
  def toast(text, duration: 3000)
    broadcast_action_later(
      :toast,
      render: false, # New in this PR
      attributes: {  # New in this PR
        text: 
        duration:
      }
    )
  end
end
```

Including it on `ApplicationRecord` so that we can `toast` to any model/streamable (as long as they're listening through `turbo_stream_from`.

```rb
class ApplicationRecord < ActiveRecord::Base
  include TurboBroadcastable
end
```

To be called anywhere, such as from a controller:

```rb
class WebhooksController < ApplicationController
  def create
    # ...
    Current.user.toast("Webhook received")
  end
end
```

## Notes

This change only affects `broadcast_action_` and related `action_` methods, since this is the method you'd use for custom actions. 

The default Javascript implementations for actions such as `prepend`, `append`, etc. are provided by Turbo on `StreamAction` and thus passing attributes here is unnecessary, since they won't be used.

### DIFF
--- a/test/streams/broadcastable_test.rb
+++ b/test/streams/broadcastable_test.rb
@@ -108,6 +108,70 @@ class Turbo::BroadcastableTest < ActionCable::Channel::TestCase
     end
   end
 
+  test "broadcasting action with attributes" do
+    assert_broadcast_on @message.to_gid_param, turbo_stream_action_tag("prepend", target: "messages", template: render(@message), "data-foo" => "bar") do
+      @message.broadcast_action "prepend", target: "messages", attributes: { "data-foo" => "bar" }
+    end
+  end
+
+  test "broadcasting action with no rendering" do
+    assert_broadcast_on @message.to_gid_param, turbo_stream_action_tag("prepend", target: "messages", template: nil) do
+      @message.broadcast_action "prepend", target: "messages", render: false
+    end
+  end
+
+  test "broadcasting action to with attributes" do
+    assert_broadcast_on "stream", turbo_stream_action_tag("prepend", target: "messages", template: render(@message), "data-foo" => "bar") do
+      @message.broadcast_action_to "stream", action: "prepend", attributes: { "data-foo" => "bar" }
+    end
+  end
+
+  test "broadcasting action to with no rendering" do
+    assert_broadcast_on "stream", turbo_stream_action_tag("prepend", target: "messages", template: nil) do
+      @message.broadcast_action_to "stream", action: "prepend", render: false
+    end
+  end
+
+  test "broadcasting action later to with attributes" do
+    @message.save!
+    
+    assert_broadcast_on @message.to_gid_param, turbo_stream_action_tag("prepend", target: "messages", template: render(@message), "data-foo" => "bar") do
+      perform_enqueued_jobs do
+        @message.broadcast_action_later_to @message, action: "prepend", target: "messages", attributes: { "data-foo" => "bar" }
+      end
+    end
+  end
+
+  test "broadcasting action later to with no rendering" do
+    @message.save!
+
+    assert_broadcast_on @message.to_gid_param, turbo_stream_action_tag("prepend", target: "messages", template: nil) do
+      perform_enqueued_jobs do
+        @message.broadcast_action_later_to @message, action: "prepend", target: "messages", render: false
+      end
+    end
+  end
+
+  test "broadcasting action later with attributes" do
+    @message.save!
+
+    assert_broadcast_on @message.to_gid_param, turbo_stream_action_tag("prepend", target: "messages", template: render(@message), "data-foo" => "bar") do
+      perform_enqueued_jobs do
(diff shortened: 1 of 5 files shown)


davidalejandroaguilar:
@marcoroth This allows calling [any of your turbo-power actions](https://github.com/marcoroth/turbo_power) from anywhere, not only from the `turbo_stream` method. e.g.

```rb
message.broadcast_action_later(
  action: :dispatch_event,
  render: false,
  attributes: {
    target: dom_id(message),
    name: "click",
    detail: {
      foo: :bar
    }
  }
)
```

Perhaps `turbo_power` could have some syntactic sugar for the above?

test/streams/broadcastable_test.rb
```diff
@@ -108,6 +108,65 @@ class Turbo::BroadcastableTest < ActionCable::Channel::TestCase
...
+
+  test "broadcasting action to with no rendering" do
+    assert_broadcast_on "stream", turbo_stream_action_tag("prepend", target: "messages", template: nil) do
+      @message.broadcast_action_to "stream", action: "prepend", render: false
+    end
+  end
+
+  # The below commented tests pass if you change `Turbo::Streams::Broadcasts#broadcast_action_later_to`'s
+  # job call from `perform_later` to `perform_now`, but for some reason `perform_enqueued_jobs` is not running them.
+
```
davidalejandroaguilar, line 137: Wonder if I should leave these tests here in case anyone can fix them. But open to just removing them altogether.

davidalejandroaguilar:
@afcapel Would you mind taking a look when you have a chance?

davidalejandroaguilar:
@seanpdoyle Hey there, it's been 4 months since I opened this PR. I know y'all are super busy, but I wonder if you could skim through this PR. It's a really simple change that creates a lot of flexibility and power. Basically it's allowing:

* Passing `attributes` to action broadcasts methods.
* Passing `render: false` to broadcast methods to avoid rendering any HTML and just sending the broadcast over the wire.

I'm unfamiliar with the repo's rules or usual wait times, so I'm unsure if this PR has been silently rejected or not. Appreciate your time and efforts spent bringing this project to life.

dhh:
I'm fine with the changes, but the test suite need to pass. Can you have a look at that?

davidalejandroaguilar:
@dhh Done, tests are passing now! Thank you.

davidalejandroaguilar:
@dhh I think I missed you by a week 🤦🏼‍♂️ , so just a friendly ping on this after 2 weeks.

Wouldn't like to pester with notifications though, so LMK if I just need to be patient.

davidalejandroaguilar:
> I'm fine with the changes, but the test suite need to pass. Can you have a look at that?

@dhh Another friendly ping, test are passing 😃

dhh:
Need to settle the commented out tests. Can't merge with that is it is.

davidalejandroaguilar:
@dhh All good now!

davidalejandroaguilar:
@dhh Another month has passed, so I'm feeling free to ping you on this 😃 

All tests are passing and ready to merge!

davidalejandroaguilar:
@dhh Pinging you on this again, hoping y'all are out of a cycle and have time to merge? 😃 

P.S. I'm pinging on a ~3-4 weeks interval so as to not annoy you. Let me know if this is too frequent.

PR #428, opened 2023-02-06
Include minimal layout in frame responses

kevinmcconnell:
Previously, responses to frame requests were rendered without any layout. Since Turbo does not need the content outside of the frame, reducing the amount that is rendered can be a useful optimisation.

However, this optimisation prevents responses from specifying `head` content as well. There are cases where it would be useful for Turbo to have access to items specified in head, like the meta tags used for specifying visit and cache control.

To get around this, we change the optimisation to use a minimal layout for frame responses, rather than no layout. With this, applications can render content into the `head` if they want, and the Turbo Rails helpers like `turbo_page_requires_reload` will also work.

### DIFF
--- a/app/controllers/turbo/frames/frame_request.rb
+++ b/app/controllers/turbo/frames/frame_request.rb
@@ -1,19 +1,27 @@
 # Turbo frame requests are requests made from within a turbo frame with the intention of replacing the content of just
 # that frame, not the whole page. They are automatically tagged as such by the Turbo Frame JavaScript, which adds a
-# <tt>Turbo-Frame</tt> header to the request. When that header is detected by the controller, we ensure that any
-# template layout is skipped (since we're only working on an in-page frame, thus can skip the weight of the layout), and
-# that the etag for the page is changed (such that a cache for a layout-less request isn't served on a normal request
-# and vice versa).
+# <tt>Turbo-Frame</tt> header to the request.
 #
-# This is merely a rendering optimization. Everything would still work just fine if we rendered everything including the layout.
-# Turbo Frames knows how to fish out the relevant frame regardless.
+# When that header is detected by the controller, we substitute our own minimal layout in place of the
+# application-supplied layout (since we're only working on an in-page frame, thus can skip the weight of the layout). We
+# use a minimal layout, rather than avoid the layout entirely, so that it's still possible to render content into the
+# <tt>head<tt>.
+#
+# Accordingly, we ensure that the etag for the page is changed, such that a cache for a minimal-layout request isn't
+# served on a normal request and vice versa.
+#
+# This is merely a rendering optimization. Everything would still work just fine if we rendered everything including the
+# full layout. Turbo Frames knows how to fish out the relevant frame regardless.
+#
+# The layout used is <tt>turbo_rails/frame.html.erb</tt>. If there's a need to customize this layout, an application can
+# supply its own (such as <tt>app/views/layouts/turbo_rails/frame.html.erb</tt>) which will be used instead.
 #
 # This module is automatically included in <tt>ActionController::Base</tt>.
 module Turbo::Frames::FrameRequest
   extend ActiveSupport::Concern
 
   included do
-    layout -> { false if turbo_frame_request? }
+    layout -> { "turbo_rails/frame" if turbo_frame_request? }
     etag { :frame if turbo_frame_request? }
   end
 
--- a/app/views/layouts/turbo_rails/frame.html.erb
+++ b/app/views/layouts/turbo_rails/frame.html.erb
@@ -1,0 +1,8 @@
+<html>
+  <head>
+    <%= yield :head %>
+  </head>
+  <body>
+    <%= yield %>
+  </body>
+</html>
--- a/test/dummy/app/views/trays/show.html.erb
+++ b/test/dummy/app/views/trays/show.html.erb
@@ -1,3 +1,7 @@
+<% content_for :head do %>
+  <meta name="test" content="present" />
+<% end %>
+
 <turbo-frame id="tray">
   <div>This is a tray!</div>
   <div><%= @frame_id %></div>
--- a/test/frames/frame_request_controller_test.rb
+++ b/test/frames/frame_request_controller_test.rb
@@ -1,12 +1,28 @@
 require "test_helper"
 
 class Turbo::FrameRequestControllerTest < ActionDispatch::IntegrationTest
-  test "frame requests are rendered without a layout" do
+  test "frame requests are rendered with a minimal layout" do
     get tray_path(id: 1)
     assert_select "title", count: 1
 
     get tray_path(id: 1), headers: { "Turbo-Frame" => "true" }
     assert_select "title", count: 0
+  end
+
+  test "frame request layout includes `head` content" do
+    get tray_path(id: 1), headers: { "Turbo-Frame" => "true" }
+
+    assert_select "head", count: 1
+    assert_select "meta[name=test][content=present]"
+  end
+
+  test "frame request layout can be overridden" do
+    with_prepended_view_path "test/frames/views" do
+      get tray_path(id: 1), headers: { "Turbo-Frame" => "true" }
+    end
+
+    assert_select "meta[name=test][content=present]"
+    assert_select "meta[name=alternative][content=present]"
   end
 
   test "frame requests get a unique etag" do
@@ -28,4 +44,13 @@ class Turbo::FrameRequestControllerTest < ActionDispatch::IntegrationTest
     get tray_path(id: 1), headers: { "Turbo-Frame" => turbo_frame_request_id }
     assert_match /#{turbo_frame_request_id}/, @response.body
   end
+
+  private
+    def with_prepended_view_path(path, &block)
+      previous_view_paths = ApplicationController.view_paths
+      ApplicationController.prepend_view_path path
+      yield
+    ensure
+      ApplicationController.view_paths = previous_view_paths
+    end
 end
--- a/test/frames/views/layouts/turbo_rails/frame.html.erb
+++ b/test/frames/views/layouts/turbo_rails/frame.html.erb
@@ -1,0 +1,9 @@
+<html>
+  <head>
+    <meta name="alternative" content="present" />
+    <%= yield :head %>
+  </head>
+  <body>
+    <%= yield %>
+  </body>
+</html>


seanpdoyle:
While I think this is a helpful change to support ongoing `@hotwired/turbo` development, I think it's worth considering the changes proposed in https://github.com/hotwired/turbo-rails/pull/232.

Conceptually, a new and separate minimized page layout provides an additional source of deviation and drift when weighed against rendering frame responses in the exact same way as other responses.

In the past, `@hotwired/turbo` has worked-around the fact that it receives incomplete pages for the Turbo-Frame response [by synthesizing a full document out of the page's `outerHTML`](https://github.com/hotwired/turbo/blob/b8ecc46d93816195898e655ebb9d3337abc98cce/src/core/frames/frame_controller.ts#L388). A change like this would affect that process. If Turbo could expect full HTML documents for `text/html` responses, that work around wouldn't be necessary.

kevinmcconnell:
I think the advantage to using a minimal layout wrapper, as opposed to a full layout, is that we'd still get the benefits of the optimisation, including the reduced rendering effort. And the small responses are easy to read when debugging, which is minor, but quite handy. The extra html/head elements here don't really impact that.

If there's a case for sending full-layout responses to frame requests, maybe we should provide people with an easy & obvious way to turn the optimisation off? (You can already do that by specifying an explicit layout, but we could make it more explicit).

seanpdoyle:
> I think the advantage to using a minimal layout wrapper, as opposed to a full layout, is that we'd still get the benefits of the optimisation, including the reduced rendering effort.
>
> If there's a case for sending full-layout responses to frame requests, maybe we should provide people with an easy & obvious way to turn the optimisation off? 

What do you think about implementing the reverse. Full responses by default, then a turn-key configuration to render with the optimized minimal layout?

app/views/layouts/turbo_rails/frame.html.erb
```diff
@@ -0,0 +1 @@
+<html><head><%= yield :head %></head><body><%= yield %></body></html>
```
seanpdoyle, line 1: If we're going to be introducing this template, what do you think about formatting its source with line breaks for the sake of legibility? I think the savings on omitting the newline character might not be pulling their weight.

```erb
<html>
  <head><%= yield :head %></head>
  <body><%= yield %>
</html>
```

kevinmcconnell: Yeah, that's a good idea 👍

app/controllers/turbo/frames/frame_request.rb
```diff
@@ -1,19 +1,24 @@
...
 # that frame, not the whole page. They are automatically tagged as such by the Turbo Frame JavaScript, which adds a
-# <tt>Turbo-Frame</tt> header to the request. When that header is detected by the controller, we ensure that any
-# template layout is skipped (since we're only working on an in-page frame, thus can skip the weight of the layout), and
-# that the etag for the page is changed (such that a cache for a layout-less request isn't served on a normal request
-# and vice versa).
+# <tt>Turbo-Frame</tt> header to the request.
 #
-# This is merely a rendering optimization. Everything would still work just fine if we rendered everything including the layout.
-# Turbo Frames knows how to fish out the relevant frame regardless.
+# When that header is detected by the controller, we substitute our own minimal layout in place of the
```
afcapel, line 5: What about mentioning the minimal layout is `turbo_rails/frame.html.erb`? People can still override the default if the want creating their own `app/views/layouts/turbo_rails/frame.html.erb` file. Probably not a very common case, but still doable if someone wants to do it.

afcapel (commented):
@kevinmcconnell looks great 🙌 

> What do you think about implementing the reverse. Full responses by default, then a turn-key configuration to render with the optimized minimal layout?

@seanpdoyle I think minimal layout by default makes more sense for two reasons:

1. It's closer to the original behavior, so less likely to introduce surprises when people upgrade.
2. Rendering a `turbo_rails/frame` also adds an extension point without any need for extra configuration. If you'd rather render the application layout you can create an `app/views/layouts/turbo_rails/frame.html.erb` and render `application.html.erb`(or whatever) from there. This would not work the other way araound. Changing the minimal response is a fairly edge case, so I think it's ok for the option to be available but not very prominent. Most people won't need any changes and they'll get optimized version out of the box.

afcapel (approved):

seanpdoyle:
@afcapel that's a great point!

> If you'd rather render the application layout you can create an app/views/layouts/turbo_rails/frame.html.erb and render application.html.erb(or whatever) from there.

Is the idea that you'd declare something like:

```erb
<%# app/views/layouts/turbo_rails/frame.html.erb %>
<%= render "layouts/application" do %>
  <%= yield %>
<% end %>
```

If that's the case, could we also introduce test coverage of some kind that exercises that behavior to make sure it's possible?

packagethief (approved):
Thanks @kevinmcconnell. I think this strikes a good balance. It ensures that frame responses contain a full HTML document, but it avoids any performance penalty from rendering a potentially-expensive application layout.

I wanted to add that I agree with @seanpdoyle's arguments in https://github.com/hotwired/turbo-rails/pull/232 and elsewhere about the benefits of returning full documents instead of fragments. I think the problem of "breaking out" of frames is a perfect illustration: as soon as frame responses are documents, things become simpler. Earlier breakout proposals are often working around the fact that there was no document to configure, and the suitability of `turbo-visit-control` was never apparent.

Hotwire's premise is that shipping HTML over the wire is fine, and that dropping down to fragments is an unnecessary optimization that in the worst case can lead to the invention of an ad-hoc, informally-specified, bug-ridden, slow implementation of half of a document format. Best to steer clear 😅

app/controllers/turbo/frames/frame_request.rb
packagethief: > What about mentioning the minimal layout is `turbo_rails/frame.html.erb`? People can still override the default if the want creating their own `app/views/layouts/turbo_rails/frame.html.erb` file.

I like that idea. Action Text's documentation makes a similar suggestion for folks needing to customize the default layout in https://guides.rubyonrails.org/action_text_overview.html#rendering-rich-text-content

kevinmcconnell:
Thanks for the feedback @seanpdoyle, @afcapel & @packagethief!

I agree that we should document the way to extend that default template. I don't think it's be something that people should need often, but we should make it clear for any cases where they do. I've added an explanation for that, and a corresponding test.

davidalejandroaguilar:
@kevinmcconnell Maybe I'm missing something, but is this not going to cause some scripts to be loaded twice?

kevinmcconnell:
@davidalejandroaguilar I'm not sure what you mean. Can you elaborate a bit?

davidalejandroaguilar:
@kevinmcconnell So I'm thinking that if you're showing `messages/index.html.erb` that has `<% content_for :head %>` that adds some scripts to your layout (which has `yield :head`). If a turbo frame response from this page also has `yield :head` it would cause the browser to load the same scripts twice.

davidalejandroaguilar:
@kevinmcconnell Thanks, I do understand that. Maybe I'm not understanding this PR. I thought this PR would cause Turbo to start including `<head>` content when placing the frame content into the DOM.

But I think I see now that the overall "ignore-everything-outside-the-turbo-frame" behavior is done on Turbo Javascript and will still keep this behavior. This PR will just add the head tag so that Turbo Javascript can pick specific information encoded there to alter some behavior, if needed.

Did I get that right?

rromanchuk:
There's something anti-pattern i must be doing, but this release broke every regular  GET link I have 

For example GET /users to GET /users/:id

> The response (200) did not contain the expected <turbo-frame id="users"> and will be ignored. To perform a full page visit instead, set turbo-visit-control to reload.

I guess because of this frame in the layout?  Don't really understand how now rendering the layout would still cause that error message

```erb
<%#  app/views/layouts/application.html.erb" %>
<body>
<%= yield %>
<%#  render "application/current_user" %>
<%= turbo_frame_tag :current_user do %>

<% end %>
</body>
```

packagethief:
> did not contain the expected <turbo-frame id="users">

This suggests the request was made from within a `turbo-frame` element with the id "users". Could there be an outer frame wrapping your links that's gone unnoticed? In previous releases this would have appeared to work, and the fact that the request originated from a frame would be ignored.

Aeramor:
This changes default behavior by including a head element. Now we have to render with `layout: false` to get the same result as before when lazy loading panels.

From docs:
```You can ensure visits to a certain page will always trigger a full reload by including a <meta name="turbo-visit-control"> element in the page’s <head>.```

turgs:
> Hi @nickjj this change doesn't affect how Turbo processes frame responses. Turbo still excludes everything outside of the frame itself, except for specific items like Turbo-specific meta tags that it checks for. So although this change sounds a bit like what you're describing, it's not really the same thing, unfortunately.

Thanks for your work on this with commits and merging it in @seanpdoyle @kevinmcconnell 

I really wish I could update the `<title>` tag with `<head>` from a turbo frame what was loaded via `data-turbo-action="advance"`. That updates the URL, but in these "advance" instances I need it to update the titlebar too. 

Seems like a few others were expecting this too:
- https://discuss.hotwired.dev/t/how-do-you-update-a-pages-title-meta-description-open-graph-and-other-head-tags-when-using-turbo-frames-with-advance/4857
- https://discuss.hotwired.dev/t/turbo-frame-with-an-advance-action-does-not-replace-page-title/3982

loust333:
Hi guys, 

Thanks for all your efforts to make turbo-rails a better gem. 

Just wanted to give my feedback concerning this pull request and @turgs comment:

In our Saas Project, we implemented turbo-rails to make the loading of the page faster, and we wanted to make the interface loading quicker than never using turbo frames. 

**Example:** 

User has a controller companies controller. From companies#show ↔ companies#edit, we added a turbo_frame with the "advance" action, as mentioned by @turgs above.  

- We saw the same problem as @turgs. The <head> tag is not updated, which means for example the <title> tag cannot be updated. In this case, it is not the worst case, our clients will have the same <title> tag which is ok 🤷‍♂️

Everything went well except after deployment on the staging. 🦸

In companies#edit we have a form which uses a "file_field" input to upload a file. Normally, there is no issue **EXCEPT** if you use the smart solution `<%= f.file_field :logo, direct_upload: true %>`  

_I use Chrome most of the time (like 98.43%) which protected me from reality. 😂_ 
  
Chrome uses cache to store the user session so if we have let's say in the <head> tag the CSRF token with   `<%= csrf_meta_tags %>` for security, we use turbo frames to load the page partially AND we have a file_field with direct_upload than Chrome will keep in mind the CSRF token when loading the turbo frame. 

The problem is that some users love Safari Browser. If we simulate the same scenario with Safari, the file_field will block the file upload AND the form with the message: `can't verify csrf token authenticity`

After an entire day of patience and lots of coffee, I understood (I think) the problem AND I found this pull request. After testing the minimal layout and failing to make it work the way I wanted, I came to the same conclusion as @turgs and others. 
If there would be a way to update a section of the <head> tag using the turbo frame and the action "advance" that would be awesome. 

Hoping this informs other people about my mistakes.

Currently, there are three SOLUTION for the issue with Safari:

- Remove turbo frame tag ✅
- Use turbo frame tag but load the entire page with target="_top" 🤷‍♂️
- Remove direct_upload: true from file_field 💔

sc0ttman:
> I was hoping this PR may have fixed it but I tested it in an app I'm building and the title along with other `<head>` tags do not get updated.

Yea it looks like you need to use Turbo Drive for this. We had disabled it by default and were hoping the same `<head>` merging functionality would exist with Turbo Frames but it appears it does not. 
If you have Drive off by default you can replace the Frame tag with a `<div data-turbo="true">` and it should behave similar with the exception that you now have to replace the full page.

nickjj:
@sc0ttman I am using drive for most things but I have a very heavy sidebar with unique user content that's not really cache'able that I wanted to avoid re-rendering by loading the main content area in with a frame when sidebar links were clicked.

I hope one day they allow this behavior with frames, it's a really common feature in my opinion to be able to update the page's title or meta tags while still using frames without having to resort to a custom hand rolled Javascript approach.

PR #429, opened 2023-02-07
Add `escape_turbo_frame` controller method

kevinmcconnell:
Add `escape_turbo_frame` controller method
    
To simplify the cases where responses might need to break out of frames, this adds an `escape_turbo_frame` method that can be used to selectively apply the reload requirement when responding to a frame request:

```ruby
class SessionsController < ApplicationController
  escape_turbo_frame only: :new
end
```


kevinmcconnell:
On reflection (and after seeing what's needed here) I'm not sure this is really worth doing after all. It's an edge case, and is already somewhat covered by using `turbo_page_requires_reload`.

PR #430, opened 2023-02-09
Add connected state to `turbo-cable-stream-source`

kevinmcconnell:
When a `turbo-cable-stream-source` is added to a page, it asynchronously opens a subscription to the appropriate stream. However it did not provide a way to programmatically tell if/when that subscription was connected.

This makes it hard to write reliable tests that depend on streams being ready for use. Or to expose that connection state in the UI (for example, to show a warning when connection is lost).

To address this, we can add a `connected` attribute to the `turbo-cable-stream-source` element whenever it is connected.

Using this, we can fix the flakiness in our broadcast tests, as now they can reliably wait for a `turbo-cable-stream-source[connected]` to be on the page before attempting to broadcast anything.


app/javascript/turbo/cable_stream_source_element.js
```diff
@@ -5,7 +5,11 @@ import snakeize from "./snakeize"
 class TurboCableStreamSourceElement extends HTMLElement {
   async connectedCallback() {
     connectStreamSource(this)
-    this.subscription = await subscribeTo(this.channel, { received: this.dispatchMessageEvent.bind(this) })
+    this.subscription = await subscribeTo(this.channel, {
+      received: this.dispatchMessageEvent.bind(this),
+      connected: this.subscriptionConnected.bind(this),
+      disconnected: this.subscriptionDisconnected.bind(this),
```
afcapel, line 11: Not for this PR, but we might be able to use fat arrows instead of `bind(this)`. Caveat is that [public class fields are only available in Safari > 14.1](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields).

afcapel (approved):
Looks great :shipit:

app/assets/javascripts/turbo.js
```diff
@@ -4061,6 +4063,12 @@ class TurboCableStreamSourceElement extends HTMLElement {
     });
     return this.dispatchEvent(event);
   }
+  subscriptionConnected() {
+    this.setAttribute("connected", "");
```
afcapel, line 4067: ✂️ semicolon

app/javascript/turbo/cable_stream_source_element.js
afcapel: BTW, no extra comma at the end of the object literal.

app/assets/javascripts/turbo.js
marcoroth, line 4067: Would it make sense to add either add a timestamp or date to the attribute value, so that you could use that value to know when it was "[last] connected"?
```suggestion
    this.setAttribute("connected", new Date());
```
or
```suggestion
    this.setAttribute("connected", Number(new Date()));
```

app/javascript/turbo/cable_stream_source_element.js
marcoroth: I don't think these are "public class fields", they are just regular methods on the class

afcapel: @marcoroth I mean that you can direclt define an event handler bound to this within a class using a fat arrow:

```js
class EventHandler {
  doSomething = (event) => {
    ...
  }
}
```

But that's implicitly using the public class fields feature.

marcoroth: @afcapel ah, I see what you mean, I misunderstood! Yeah, that's right 👍🏼

app/assets/javascripts/turbo.js
kevinmcconnell: I'm inclined to leave it as-is, because I prefer the simplicity of a boolean attribute here.

That said, I also can't really think of a case where the timestamp would be needed. I'm only interested in the current state. Did you have a specific case in mind where you would need it?

PR #431, opened 2023-02-09
Use Rack version < 3 on edge Rails

kevinmcconnell:
There is some test failure when running against edge Rails, due to the dependencies installing Rack 3 which is [not yet supported][1]. Pinning our Rack version prevents the failures.

Once the linked Rails PR has landed, we can remove this constraint.

[1]: https://github.com/rails/rails/pull/46594

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,10 @@ gem 'rake'
 gem 'byebug'
 gem 'puma'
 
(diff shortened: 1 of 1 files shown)


kevinmcconnell:
@afcapel this is the last step to get CI green again \o/

afcapel (approved):
Nice catch 🚀

PR #433, opened 2023-02-12
Support renderable object as turbo stream content

joeldrapper:
Add support for _renderable_ objects as turbo stream content. Renderable objects respond to `render_in(view_context, &block)`, returning an html-safe String (an `ActiveSupport::SafeBuffer`).

### Example

```ruby
turbo_stream.append "notifications", NotificationComponent.new(@notification)
```

This will improve compatibility with [Phlex](https://www.phlex.fun), [ViewComponent](https://viewcomponent.org), and anything else that uses the _renderable_ duck type to provide object-oriented views.

### Approach

I added a case to `render_template` in the `TagBuilder` for when the `content` argument is a _renderable_, based on whether it responds to `render_in`.

I’m not sure if this was the right testing strategy — I decided to make a new example resource ("notifications") rather than modify one of the existing examples. Please let me know if you have any suggestions on how to better approach that.

### DIFF
--- a/app/models/turbo/streams/tag_builder.rb
+++ b/app/models/turbo/streams/tag_builder.rb
@@ -227,6 +227,8 @@ class Turbo::Streams::TagBuilder
   private
     def render_template(target, content = nil, allow_inferred_rendering: true, **rendering, &block)
       case
+      when content.respond_to?(:render_in)
+        content.render_in(@view_context, &block)
       when content
         allow_inferred_rendering ? (render_record(content) || content) : content
       when block_given?
--- a/test/dummy/app/components/notification_component.rb
+++ b/test/dummy/app/components/notification_component.rb
@@ -1,0 +1,5 @@
+class NotificationComponent
+  def render_in(view_context, &block)
+    "<div class='notification'>#{view_context.capture(&block)}</div>".html_safe
+  end
+end
--- a/test/dummy/app/controllers/notifications_controller.rb
+++ b/test/dummy/app/controllers/notifications_controller.rb
@@ -1,0 +1,7 @@
+class NotificationsController < ApplicationController
+  def create
+    respond_to do |format|
+      format.turbo_stream { render :new }
+    end
+  end
+end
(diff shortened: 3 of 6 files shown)


app/models/turbo/streams/tag_builder.rb
```diff
@@ -227,6 +227,8 @@ def action_all(name, targets, content = nil, allow_inferred_rendering: true, **r
   private
     def render_template(target, content = nil, allow_inferred_rendering: true, **rendering, &block)
       case
+      when content.respond_to?(:render_in)
+        content.render_in(@view_context, &block)
```
joeldrapper, line 231: This line could instead be `@view_context.render(content, &block)`, which would give more control to the view context but will be slightly slower because of the extra indirection.

We’re already coupled to the `render_in` method based on the `respond_to?` check in the previous line, though that too could be delegated to the view context if the view context exposes a renderable-checking method.

I personally think this coupling with the `render_in` interface is _fine_, but open to your input and wanted to highlight an alternative.

seanpdoyle: My preference is to delegate to Action View as much as possible, going as far as invoking `render` when we know that it'll call `render_in` behind the scenes.

seanpdoyle (approved):
Thank you for taking this one on!

app/models/turbo/streams/tag_builder.rb
joeldrapper: That makes sense. The reason I didn’t do that was because we’re already coupled to `render_in` here via the `respond_to?` check. Shall I update it to this?

```ruby
when content.respond_to?(:render_in)
  @view_context.render(content, &block)
```

joelhawksley:
👋🏻 @seanpdoyle do you know who would be best to ping for a review to get this PR in? I'd love to see this land for folks using ViewComponent and Turbo like us at GitHub 😄

seanpdoyle:
@afcapel @kevinmcconnell, maybe?

kevinmcconnell (approved):
This looks great, thanks @joeldrapper!

And thanks for reviewing, @seanpdoyle.

Issue #340, opened 2022-05-26
After updated to 1.1.1, ActionText attachments are not showing


YutoYasunaga:
After updated turbo-rails to version 1.1.1. persisted ActionText attachments are not showing anymore, it looks like this:

![image](https://user-images.githubusercontent.com/6220668/170416938-4c1a0509-ff4b-4340-a3e4-c5d0eb4dde2e.png)

**When I restored to version 1.0.1, it works fine.**

dhh:
Can't replicate that on a new Rails app with turbo-rails 1.1.1 and Action Text. Need more detail. Look in the logs, look in the JavaScript console for errors.

artisr:
I also have the same issue 
This https://github.com/hotwired/turbo-rails/commit/102e919318ec528b9f5908e10d6427847d7a197b is to blame, at least in my case

ccasabona:
I have the same problem on an app with around 125 attachments.  

Here is a link to a demo app, as requested above.   The demo uses Rails 7.0.3, and turbo-rails with importmaps.  This was setup using ruby 3.1.2.

To demonstrate the problem, first use the app and attach an image using turbo-rails 1.0.1.  Then upgrade the v 1.1.1.  The image does does not display.   I get 404 errors in the console and in the rails log.

 If you attach the same image using turbo-rails 1.1.1, it displays properly.  This is impractical for apps with many attachments.

[https://github.com/ccasabona/noatt.git]( https://github.com/ccasabona/noatt.git)

tleish:
@ccasabona - I just tried the demo app on ruby 3.1.2.  It worked for me on turbo-rails 1.0.1 and then again on and turbo-rails 1.1.1.  I am unable to reproduce the issue.  I tested on MacOS using the latest Chrome and Firefox, worked on both.  Note: I do have "disable caching" in the browser enabled.  Is it possible there might be a cached javascript asset?

ccasabona:
Still the same problem with caching disabled on latest Safari, Firefox, and Chrome using MacOS.  I notice in the rails log that after upgrading to 1.1.1, the ActiveStorage::Blob Load statement is not present.

Here is my log under v 1.0.1:

```
Started GET "/contents/1" for ::1 at 2022-06-02 07:33:31 -0400
Processing by ContentsController#show as HTML
  Parameters: {"id"=>"1"}
[...]

PR #436, opened 2023-02-22
Document solution for issue #340

kevinmcconnell:
Applications that are affected by secrets changing when upgrading pass 1.1.0 can use a key rotation to avoid ActiveStorage assets breaking. This adds documentation to explain the issue, with code showing how to rotate the key.


afcapel (approved):
Looks great 👍

PR #438, opened 2023-03-03
Document Navigation and expose methods publicly

joemasilotti:
This PR exposes the Turbo Native navigation helpers to be used in developer's apps. It also add some documentation on how you might want to use them.

### DIFF
--- a/app/controllers/turbo/native/navigation.rb
+++ b/app/controllers/turbo/native/navigation.rb
@@ -2,7 +2,7 @@
 # have Turbo Native clients running (see the Turbo iOS and Turbo Android projects for details), you can respond to native
 # requests with three dedicated responses: <tt>recede</tt>, <tt>resume</tt>, <tt>refresh</tt>.
 #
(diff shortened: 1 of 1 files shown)


app/controllers/turbo/native/navigation.rb
```diff
@@ -2,23 +2,28 @@
 # have Turbo Native clients running (see the Turbo iOS and Turbo Android projects for details), you can respond to native
 # requests with three dedicated responses: <tt>recede</tt>, <tt>resume</tt>, <tt>refresh</tt>.
 #
-# FIXME: Supply full description of when we use either.
+# turbo-android handles these actions automatically. You are require to implement the handling on your own for turbo-ios.
```
jordelver, line 5:
```suggestion
# turbo-android handles these actions automatically. You are required to implement the handling on your own for turbo-ios.
```

joemasilotti: Good catch, thanks!

PR #439, opened 2023-03-08
feat: add action attribute to turbo_frame helper

benkoshy:
### What is this?

Developer convenience.

### Why do I need this? 

Right now I have to type out:

```rb
turbo_frame_tag("message", src: record, "data_turbo_action": "advance")
```
....wait, did you catch the error above? 

Why should we trouble ourselves with remembering whether something is snake case or kebab case? Or waste time debugging, checking the docs?

This is a minor case of [computational compression](https://m.signalvnoise.com/conceptual-compression-means-beginners-dont-need-to-know-sql-hallelujah/).

Just pass in the action directly, and let's blaze along:

```rb
turbo_frame_tag("message", src: record, action: "advance")
```

### DIFF
--- a/app/helpers/turbo/frames_helper.rb
+++ b/app/helpers/turbo/frames_helper.rb
@@ -35,10 +35,10 @@ module Turbo::FramesHelper
   #
   #   <%= turbo_frame_tag(Article.find(1), Comment.new) %>
   #   # => <turbo-frame id="article_1_new_comment"></turbo-frame>
-  def turbo_frame_tag(*ids, src: nil, target: nil, **attributes, &block)
+  def turbo_frame_tag(*ids, src: nil, target: nil, action: nil, **attributes, &block)
     id = ids.map { |id| id.respond_to?(:to_key) ? ActionView::RecordIdentifier.dom_id(id) : id }.join("_")
     src = url_for(src) if src.present?
 
-    tag.turbo_frame(**attributes.merge(id: id, src: src, target: target).compact, &block)
+    tag.turbo_frame(**attributes.merge(id: id, src: src, target: target, "data-turbo-action": action).compact, &block)
   end
 end
--- a/test/frames/frames_helper_test.rb
+++ b/test/frames/frames_helper_test.rb
@@ -37,4 +37,10 @@ class Turbo::FramesHelperTest < ActionView::TestCase
   test "block style" do
     assert_dom_equal(%(<turbo-frame id="tray"><p>tray!</p></turbo-frame>), turbo_frame_tag("tray") { tag.p("tray!") })
   end
+
+  test "frame with action" do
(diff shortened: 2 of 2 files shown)


pinzonjulian:
I was bit by this a few weeks ago and thought of making this very same PR. This does feel very intuitive and in a very ruby fashion, the method or signature that you thought existed, exists.

However,

All of rails' `_tag` helpers represent an HTML element. In this case, the `<turbo-frame>`  custom element.
[This element does not currently have](https://turbo.hotwired.dev/reference/frames#frame-that-promotes-navigations-to-visits) an `action` attribute but rather uses the [`data-turbo-action="[advance | replace]"`](https://turbo.hotwired.dev/reference/attributes)  one.

This is due to the fact that it's a shared attribute between `<turbo-frame>` and `<a>` tags

When I realised this, I remembered that I use `data` attributes in `_tag` methods and other view helpers by treating them like a hash:

```ruby
link_to @user, data: { turbo_action: "replace" }
form_with model: @user, data: { turbo_frame: "drawer" } do |form| 
  # ...
end
```

Maybe it would be good to document this rather than change the overall convention set in every other html tag and helper method.

benkoshy:
@pinzonjulian I thank you for your thoughts. The argument presented:

> Maybe it would be good to document this rather than change the overall convention set in every other html tag and helper method.

In other words, we already have a [knife](https://rubyonrails.org/doctrine#provide-sharp-knives) for this particular situation (?). 

```rb
# option 1
turbo_frame_tag("message", src: record, data: {turbo_action: "advance"}) # kebab-case doesn't matter
```

But it's a blunt knife. Why don't we make it sharper? We get this for free (big assumption):

```rb
# option 2
turbo_frame_tag("message", src: record, action: "advance")
# superfluous: "data" and "turbo"
```

Personally, I would prefer option 2. 

Thoughts?

pinzonjulian:
I kind of agree with you that it feels better. But there is a battle here between feeling good and breaking conventions. 

The `id`, `src`, `target`and `loading` attributes are native attributes of the `<turbo-frame>` custom element so it makes sense for them to be arguments of the `turbo_frame_tag` method.

Making `action` an available argument makes it seem that it's an attribute of the frame, which it is not. 

I'm just stating the facts and the way Rails has set the conventions for these types of view helpers. 

I'm happy to hear the maintainers' and authors' point of view!

marcoroth:
I think I agree with @pinzonjulian here. Since `turbo_frame_tag` is just a Rails tag helper it's kinda weird to have some implicit logic to convert attributes just for `<turbo-frame>` elements.

That's why I personally would also prefer option 1 from above, because that way it's consistent across the board, no matter on which element the `data-turbo-action` attribute is placed.
```ruby
turbo_frame_tag("message", src: record, data: { turbo_action: "advance" })
```

PR #465, opened 2023-05-11
[CI] - gemfile - allow for capybara options

MSP-Greg:
Puma runs CI against the main branch here, as a 'real world' test.  I think a change in selenium-webdriver broke newer versions of capybara, which are used with Ruby 3.x.  Hence, CI broke here and in Puma.

The issue is fixed in capybara, but the fix hasn't been released.  This PR selectively allows using capybara from the repo if needed.

When capybara does a new release, `{git: 'https://github.com/teamcapybara/capybara.git', ref: '43e32a8495'}` can be replaced with `{}`.

Note that gemfiles cannot list the same gem twice, which is the reason for the logic with 'kw'.

Not sure if you're interested in this...

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,7 +26,13 @@ group :development, :test do
 end
 
 group :test do
-  gem 'capybara'
+  kw =
+    if RUBY_VERSION.start_with? '3'
+      {git: 'https://github.com/teamcapybara/capybara.git', ref: '43e32a8495'}
+    else
+      {}
+    end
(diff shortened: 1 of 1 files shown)


MSP-Greg:
Capybara released 3.39.1, which fixes the failing CI.  I believe their change was needed due to a change in selenium-webdriver.

Anyway, I'll close this.  In Puma, we added code similar to this, but commented it out when 3.3.9.1 was released.

PR #466, opened 2023-05-13
Introduce `Turbo::Broadcastable::TestHelper`

seanpdoyle:
The `Turbo::Broadcastable::TestHelper` concern provides Action
Cable-aware test helpers that assert that `<turbo-stream>` elements were
or were not broadcast over Action Cable. They are not automatically
included.

`assert_turbo_stream_broadcasts`
---

Asserts that `<turbo-stream>` elements were broadcast over Action Cable

**Arguments**

* `stream_name_or_object` the objects used to generate the
  channel Action Cable name, or the name itself
* `&block` optional block executed before the
  assertion

**Options**

* `count:` the number of `<turbo-stream>` elements that are
expected to be broadcast

Asserts `<turbo-stream>` elements were broadcast:

```ruby
message = Message.find(1)
message.broadcast_replace_to "messages"

assert_turbo_stream_broadcasts "messages"
```

Asserts that two `<turbo-stream>` elements were broadcast:

```ruby
message = Message.find(1)
message.broadcast_replace_to "messages"
message.broadcast_remove_to "messages"

assert_turbo_stream_broadcasts "messages", count: 2
```

You can pass a block to run before the assertion:

```ruby
message = Message.find(1)

assert_turbo_stream_broadcasts "messages" do
  message.broadcast_append_to "messages"
end
```

In addition to a String, the helper also accepts an Object or Array to
determine the name of the channel the elements are broadcast to:

```ruby
message = Message.find(1)

assert_turbo_stream_broadcasts message do
  message.broadcast_replace
end
```

`assert_no_turbo_stream_broadcasts`
---

Asserts that no `<turbo-stream>` elements were broadcast over Action Cable

**Arguments**

* `stream_name_or_object` the objects used to generate the
  channel Action Cable name, or the name itself
* `&block` optional block executed before the
  assertion

Asserts that no `<turbo-stream>` elements were broadcast:

```ruby
message = Message.find(1)
message.broadcast_replace_to "messages"

assert_no_turbo_stream_broadcasts "messages" # fails with MiniTest::Assertion error
```

You can pass a block to run before the assertion:

```ruby
message = Message.find(1)

assert_no_turbo_stream_broadcasts "messages" do
  # do something other than broadcast to "messages"
end
```

In addition to a String, the helper also accepts an Object or Array to
determine the name of the channel the elements are broadcast to:

```ruby
message = Message.find(1)

assert_no_turbo_stream_broadcasts message do
  # do something other than broadcast to "message_1"
end
```

`capture_turbo_stream_broadcasts`
---

Captures any `<turbo-stream>` elements that were broadcast over Action Cable

**Arguments**

* `stream_name_or_object` the objects used to generate the
  channel Action Cable name, or the name itself
* `&block` optional block to capture broadcasts during execution

Returns any `<turbo-stream>` elements that have been broadcast as an
Array of `Nokogiri::XML::Element` instances

```ruby
message = Message.find(1)
message.broadcast_append_to "messages"
message.broadcast_prepend_to "messages"

turbo_streams = capture_turbo_stream_broadcasts "messages"

assert_equal "append", turbo_streams.first["action"]
assert_equal "prepend", turbo_streams.second["action"]
```

You can pass a block to limit the scope of the broadcasts being captured:

```ruby
message = Message.find(1)

turbo_streams = capture_turbo_stream_broadcasts "messages" do
  message.broadcast_append_to "messages"
end

assert_equal "append", turbo_streams.first["action"]
```

In addition to a String, the helper also accepts an Object or Array to
determine the name of the channel the elements are broadcast to:

```ruby
message = Message.find(1)

replace, remove = capture_turbo_stream_broadcasts message do
  message.broadcast_replace
  message.broadcast_remove
end

assert_equal "replace", replace["action"]
assert_equal "replace", remove["action"]
```

### DIFF
--- a/lib/turbo/broadcastable/test_helper.rb
+++ b/lib/turbo/broadcastable/test_helper.rb
@@ -1,0 +1,172 @@
+module Turbo
+  module Broadcastable
+    module TestHelper
+      extend ActiveSupport::Concern
+
+      included do
+        include ActionCable::TestHelper
+
+        include Turbo::Streams::StreamName
+      end
+
+      # Asserts that `<turbo-stream>` elements were broadcast over Action Cable
+      #
+      # === Arguments
+      #
+      # * <tt>stream_name_or_object</tt> the objects used to generate the
+      #   channel Action Cable name, or the name itself
+      # * <tt>&block</tt> optional block executed before the
+      #   assertion
+      #
+      # === Options
+      #
+      # * <tt>count:</tt> the number of `<turbo-stream>` elements that are
+      # expected to be broadcast
+      #
+      # Asserts `<turbo-stream>` elements were broadcast:
+      #
+      #     message = Message.find(1)
+      #     message.broadcast_replace_to "messages"
+      #
+      #     assert_turbo_stream_broadcasts "messages"
+      #
+      # Asserts that two `<turbo-stream>` elements were broadcast:
+      #
+      #     message = Message.find(1)
+      #     message.broadcast_replace_to "messages"
+      #     message.broadcast_remove_to "messages"
+      #
+      #     assert_turbo_stream_broadcasts "messages", count: 2
+      #
+      # You can pass a block to run before the assertion:
+      #
+      #     message = Message.find(1)
+      #
+      #     assert_turbo_stream_broadcasts "messages" do
+      #       message.broadcast_append_to "messages"
+      #     end
+      #
+      # In addition to a String, the helper also accepts an Object or Array to
+      # determine the name of the channel the elements are broadcast to:
+      #
+      #     message = Message.find(1)
+      #
+      #     assert_turbo_stream_broadcasts message do
+      #       message.broadcast_replace
+      #     end
+      #
+      def assert_turbo_stream_broadcasts(stream_name_or_object, count: nil, &block)
+        payloads = capture_turbo_stream_broadcasts(stream_name_or_object, &block)
+        stream_name = stream_name_from(stream_name_or_object)
+
+        if count.nil?
+          assert_not_empty payloads, "Expected at least one broadcast on #{stream_name.inspect}, but there were none"
+        else
+          broadcasts = "Turbo Stream broadcast".pluralize(count)
+
+          assert count == payloads.count, "Expected #{count} #{broadcasts} on #{stream_name.inspect}, but there were none"
+        end
+      end
+
+      # Asserts that no `<turbo-stream>` elements were broadcast over Action Cable
+      #
+      # === Arguments
+      #
+      # * <tt>stream_name_or_object</tt> the objects used to generate the
+      #   channel Action Cable name, or the name itself
+      # * <tt>&block</tt> optional block executed before the
+      #   assertion
+      #
+      # Asserts that no `<turbo-stream>` elements were broadcast:
+      #
+      #     message = Message.find(1)
+      #     message.broadcast_replace_to "messages"
+      #
+      #     assert_no_turbo_stream_broadcasts "messages" # fails with MiniTest::Assertion error
+      #
+      # You can pass a block to run before the assertion:
+      #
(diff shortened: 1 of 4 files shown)


dhh:
I like this, but find it a bit awkward that the assert is also responsible for a return. We typically don't have double meanings like that for assert. 

How about separating out the turbo stream capturing, so we have:

```ruby
actions = capture_turbo_broadcasts(stream: "messages") { message.broadcast_append_to ... }
```

And then explicit asserts:

```ruby
assert_turbo_broadcast(action: "append", stream: "messages") { message.broadcast_append_to ... }
```

seanpdoyle:
The `assert_turbo_broadcasts` signature draws direct inspiration from `assert_broadcasts`, since it's built atop Action Cable's helpers.

Action Cable's [assert_broadcasts](https://edgeapi.rubyonrails.org/classes/ActionCable/TestHelper.html#method-i-assert_broadcasts) test helper returns the broadcasts as objects when the assertion is passed a block:

```ruby
def test_broadcasts_again
  message = assert_broadcasts('messages', 1) do
    ActionCable.server.broadcast 'messages', { text: 'hello' }
  end
  assert_equal({ text: 'hello' }, message)

  messages = assert_broadcasts('messages', 2) do
    ActionCable.server.broadcast 'messages', { text: 'hi' }
    ActionCable.server.broadcast 'messages', { text: 'how are you?' }
  end
  assert_equal 2, messages.length
  assert_equal({ text: 'hi' }, messages.first)
  assert_equal({ text: 'how are you?' }, messages.last)
end
```

This style of `assert_broadcasts` was introduced in https://github.com/rails/rails/pull/47793, and has been released as part of [7.0.6](https://github.com/rails/rails/blob/v7.0.6/actioncable/lib/action_cable/test_helper.rb#L21-L56).

lib/turbo/broadcastable/test_helper.rb
```diff
@@ -0,0 +1,120 @@
...
+      #       message.broadcast_replace
+      #     end
+      #
+      #     assert_equal "replace", turbo_streams.first["action"]
+      #
+      def assert_turbo_stream_broadcasts(stream_name_or_object, &block)
+        block&.call
+
+        stream_name = stream_name_from(stream_name_or_object)
+        payloads = broadcasts(stream_name)
```
seanpdoyle, line 64: Since this uses the [broadcasts](https://edgeapi.rubyonrails.org/classes/ActionCable/SubscriptionAdapter/Test.html#method-i-broadcasts) test helper, it might be worthwhile to extract a `turbo_broadcasts` helper to invoke `broadcasts` and transform them into a collection of Nokogiri nodes. With that public method extracted, this helper could re-use it along with any other test case.

dhh:
Thanks for highlight those additions to Rails. I've made the same comments on those PRs as on this. I don't want to see assert grow side-effect behavior in any of the test helpers. But very happy to see us introduce a separate `capture_*` set of methods to do this.

ghiculescu:
> This style of `assert_broadcasts` was introduced in [rails/rails#47793](https://github.com/rails/rails/pull/47793), and has been released as part of [7.0.6](https://github.com/rails/rails/blob/v7.0.6/actioncable/lib/action_cable/test_helper.rb#L21-L56).

Unless I'm missing it, I don't think either of those PRs (https://github.com/rails/rails/pull/47025 or https://github.com/rails/rails/pull/47793) has been included in a release yet.

dhh:
Thanks for fixing these. You're right that they were not part of a release yet, so easy to change now.

ghiculescu:
thanks for the feedback!

ghiculescu:
FYI https://github.com/rails/rails/issues/48934

I think the feedback about `assert_broadcasts` is valid, but not really a bug in Rails. Raising it here in case you want to handle it in turbo-rails.

PR #470, opened 2023-06-05
Add a note about overwriting layout resolution and turbo frame requests

### LINKED ISSUE #268: Layout not omitted for turbo frames when using custom layout
Let's say having:
```
AdminController < ApplicationController
  layout :admin
end
```
When I use turbo frame request in sample controller:
```
UsersController < AdminController
  def show
  end
end
```
The optimization to omit layout in https://github.com/hotwired/turbo-rails/blob/a0f14b9e4649f60bcf8fa8b82f35e7f460d60177/app/controllers/turbo/frames/frame_request.rb#L16 does not work.

I can see that this bug is known, ie https://github.com/hotwired/turbo-rails/pull/232#issuecomment-930268246 but opening this issue to keep track on this specific problem.

cice:
Closes #268 

Hi,
since we also ran into the issue of using Turbo-Frames with custom layouts, I took the liberty to draft an extension to the Readme to cover this topic.
I am neither a native english speaker, nor an experienced tech/documentation writer, so feel free to correct/extend/rewrite.
cheers,
Marian

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,37 @@ When the user will click on the `Edit this todo` link, as direct response to thi
 
 [See documentation](https://turbo.hotwired.dev/handbook/frames).
 
+### A note on custom layouts
+
+In order to render turbo frame requests without the application layout, Turbo registers a custom [layout method](https://api.rubyonrails.org/classes/ActionView/Layouts/ClassMethods.html#method-i-layout). 
+If your application uses custom layout resolution, you have to make sure to return `false` for turbo frame requests:
+
+```ruby
+layout :custom_layout
+
+def custom_layout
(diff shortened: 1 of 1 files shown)


cice:
> Does this result in no layout being set? Isn't that different than
> 
> https://github.com/hotwired/turbo-rails/blob/a4b46200a0a676b3490c54048cd5d478c19d9b4a/app/controllers/turbo/frames/frame_request.rb#L24

That is a valid point. Apparently this was changed (more or less) recently: https://github.com/hotwired/turbo-rails/pull/428

dhh:
Wups, yes, true. @cice can you update docs to match #428?

PR #476, opened 2023-06-20
Make turbo_frame_tag fully compatible with dom_id

skipkayhil:
Ref #449 

Previously, there was an inconsistency with how turbo_frame_tag and dom_id would create ids:

```
turbo_frame_tag(Article.find(1), "comments") # => <turbo-frame id="article_1_comments" />
dom_id(Article.find(1), "comments") # => "comments_article_1"
```

This commit fixes that incompatibility by fully delegating id generation to dom_id when given a model object.

```
turbo_frame_tag(Article.find(1), "comments") # => <turbo-frame id="comments_article_1" />
dom_id(Article.find(1), "comments") # => "comments_article_1"

turbo_frame_tag("comments") # => <turbo-frame id="comments" />
```

This does break the use case of passing multiple models, but that seems like it would be better fixed in dom_id itself.


brunoprietog:
I agree with this change, but upgrading broke a lot of Turbo Frames. Don't you think we should announce it with more emphasis? Is it an option to support both ways?

PR #479, opened 2023-06-27
Make activejob optional

### LINKED ISSUE #478: active_job must be required
See https://github.com/hotwired/turbo-rails/issues/330, which was closed but never fixed

The dependency was added in https://github.com/hotwired/turbo-rails/pull/331, but this was not sufficient. ActiveJob actually has to be required to be defined. Alternatively, the turbo job could only be declared if ActiveJob is defined.

Original text:

Because of https://github.com/hotwired/turbo-rails/blob/main/app/jobs/turbo/streams/broadcast_job.rb and https://github.com/hotwired/turbo-rails/blob/main/app/jobs/turbo/streams/action_broadcast_job.rb and the eager loader, active_job needs to be required by any app that is using turbo-rails.

Should turbo-rails somehow make these optional or require active_job explicitly?

A user who does not require active_job will see an uninitialized constant ActiveJob typically when they deploy to a production-like environment for the first time.

aaronjensen:
Fixes #478 

There was already precedent for optional Rails dependencies (action cable), so I followed that.

We could consider combining the two initializers at this point, but I'm not sure what we'd name it.

### DIFF
--- a/lib/turbo/engine.rb
+++ b/lib/turbo/engine.rb
@@ -15,6 +15,10 @@ module Turbo
       #{root}/app/jobs
     )
 
+    initializer "turbo.no_active_job", before: :set_eager_load_paths do
+      config.eager_load_paths.delete("#{root}/app/jobs") unless defined?(ActiveJob)
+    end
+
     initializer "turbo.no_action_cable", before: :set_eager_load_paths do
       config.eager_load_paths.delete("#{root}/app/channels") unless defined?(ActionCable)
     end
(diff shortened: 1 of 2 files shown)


aaronjensen:
@afcapel Is this something you would be able to merge? It'd nice to get this one in. Thank you.

aaronjensen:
@seanpdoyle along the same lines as #573 and #574 if you would merge this to support not requiring ActiveJob, I'd appreciate it. Thank you.

aaronjensen:
Replaced by #602

PR #488, opened 2023-07-27
Removed unnecessary double quote in README.md

ruzia:
I noticed an unnecessary double quote, so I fixed it!


kevinmcconnell:
Well spotted! Thanks for the fix 🙏

ruzia:
✌️

PR #489, opened 2023-07-27
Update MiniTest -> Minitest

kevinmcconnell:
Minitest recently dropped compatibility with the older Pascal-cased format of the module name. So to avoid failures on recent Minitest versions we should use the newer format.


kevinmcconnell:
This is to fix the recent CI failures on `main`.

afcapel (approved):
Looks good 👍

PR #492, opened 2023-09-09
fix: typescript reference on readme

drochetti:
In the middle of the recent drama, my contribution to the discussion is: keep things accurate no matter what. ✌️❤️


afcapel:
Thanks @drochetti

PR #495, opened 2023-09-16
Add support for bun

rubys:
If you create a new rails project with "--main --javascript bun" you end up with a yarn.lock and not a bun.lockb file.  This is because turbo-rails and stimuls are installed with yarn.


afcapel:
@rubys thanks for the contribution, but that was implemented at the same time in https://github.com/hotwired/turbo-rails/pull/494

PR #499, opened 2023-10-05
Support for page refreshes and broadcasting

afcapel:
This PR is the Rails companion for https://github.com/hotwired/turbo/pull/1019

### Helper to configure page refresh directives

```ruby
turbo_refreshes_with scroll method: :morph, scroll: :preserve
```

### Helpers to broadcast page refresh actions

This adds new Active Record helpers to broadcast page refreshes from models:

```ruby
class Board
  broadcast_refreshes
end
```

This works great in hierarchical structures, where the child record touches parent records automatically to invalidate the cache:

```ruby
class Column
  belongs_to :board, touch: true # +Board+ will trigger a page refresh on column changes
end
```

You can also specify the streamable declaratively:

```ruby
class Column
  belongs_to :board
  broadcast_refreshes_to :board
end
```

There are also instance-level companion methods to broadcast page refreshes:

- `broadcast_refresh_later`
- `broadcast_refresh_later_to(*streamables)`

### System to suppress stream broadcasts

This PR introduces a new mechanism to suppress the broadcasting of turbo streams for arbitrary blocks of code:

```ruby
Recording.suppressing_turbo_broadcasts do
...
end
```

### System to debounce broadcasted page refresh jobs

When the broadcasting page refreshes, the system will automatically debounce multiple calls in a row to only broadcast the last one. This is meant for scenarios where you process records in mass. Because of the nature of such signals, it makes no sense to broadcast them repeatedly and individually.

The debounce implementation uses `concurrent-ruby` and thread variables under the hood.


jvillarejo:
Should the debouncer implementation using `concurrent-ruby` be part of `ActiveSupport` or be part of this gem ?

Also, is the debounce for multiple calls useful only for page refreshes or for the other broadcasting methods too? 
The broadcast_replace_to might be a good candidate.

northeastprince:
I would introduce the suppression system in a separate PR.

app/models/concerns/turbo/broadcastable.rb
```diff
@@ -112,10 +135,34 @@ def broadcasts(stream = model_name.plural, inserts_by: :append, target: broadcas
...
+    # Executes +block+ preventing both synchronous and asynchronous broadcasts from this model.
+    def suppressing_turbo_broadcasts(&block)
+      original, self.suppressed_turbo_broadcasts = self.suppressed_turbo_broadcasts, true
+      yield
+    ensure
+      self.suppressed_turbo_broadcasts = original
+    end
+
+    def suppressed_turbo_broadcasts?
+      suppressed_turbo_broadcasts
```
northeastprince, line 164:
```suggestion
      !!suppressed_turbo_broadcasts
```

app/models/concerns/turbo/broadcastable.rb
```diff
@@ -112,10 +135,34 @@ def broadcasts(stream = model_name.plural, inserts_by: :append, target: broadcas
...
+    # Configures the model to broadcast a "page refresh" on creates, updates, and destroys to a stream
+    # name derived at runtime by the <tt>stream</tt> symbol invocation.
+    def broadcasts_refreshes_to(stream)
+      after_commit -> { broadcast_refresh_later_to(stream.try(:call, self) || send(stream)) }
+    end
+
+    # Same as <tt>#broadcasts_refreshes_to</tt>, but the designated stream for page refreshes is automatically set to
+    # the current model.
+    def broadcasts_refreshes
+      after_commit -> { broadcast_refresh_later }
```
tonysm, line 147: So, this will only send broadcast refreshes to the current model's channel, right? Shouldn't it be closer to the `broadcasts` method and accept the `stream` argument or default to the model's plural name when creating the model? Otherwise, it will send a broadcast to a channel that no one will ever be listening on (because the model is being created), right?

tonysm, line 147: So, this will only send broadcast refreshes to the current model's channel, right? Shouldn't it be closer to the broadcasts method and accept the stream argument or default to the model's plural name when creating the model? Otherwise, it will send a broadcast to a channel that no one will ever be listening on (because the model is being created), right?
```suggestion
    def broadcasts_refreshes(stream = model_name.plural)
      after_create_commit -> { broadcast_refresh_later_to(stream) }
      after_update_commit -> { broadcast_refresh_later }
```

tonysm: Sent a PR: https://github.com/hotwired/turbo-rails/pull/521

jon-sully:
Just out of curiosity, why go with the `_later` approach for the broadcasts with refreshes. @afcapel raised a key point to my (admittedly "duh") [question](https://github.com/hotwired/turbo-rails/issues/506) the other month in that we don't have a `broadcast_remove_later` method since broadcasting a 'remove' action requires no markup, and thus no background processing to save markup-time is necessary. Wouldn't this be the same, since there's no real markup, just a single (constant) [html tag](https://github.com/hotwired/turbo-rails/pull/499/files#diff-a1d22ee1a2d45d29235e38916610bbd54f19d4df1be137ca2686469d5cf60a09R73)?

PR #501, opened 2023-10-13
Allow broadcasting a renderable later

davidalejandroaguilar:
# Description

https://github.com/hotwired/turbo-rails/pull/364 added support for passing a `renderable` to `Broadcastable` methods. However, it only added it for the synchronous methods. The asynchronous (`_later`) methods are broken because the renderable instance is not serialized.

This PR serializes the renderable before enqueueing it.

### DIFF
--- a/app/channels/turbo/streams/broadcasts.rb
+++ b/app/channels/turbo/streams/broadcasts.rb
@@ -64,6 +64,11 @@ module Turbo::Streams::Broadcasts
   end
 
   def broadcast_action_later_to(*streamables, action:, target: nil, targets: nil, **rendering)
+    # Serialize the renderable object if it is present.
+    if rendering[:renderable]
+      rendering[:html] = rendering.delete(:renderable).render_in(self)
+    end
+
     Turbo::Streams::ActionBroadcastJob.perform_later \
       stream_name_from(streamables), action: action, target: target, targets: targets, **rendering
   end
--- a/test/system/broadcasts_test.rb
+++ b/test/system/broadcasts_test.rb
@@ -37,6 +37,28 @@ class BroadcastsTest < ApplicationSystemTestCase
     assert_selector("title", count: 1, visible: false, text: "Dummy")
   end
 
+  test "Message broadcast append later with renderable: render option" do
+    visit messages_path
+    wait_for_stream_to_be_connected
+  
+    perform_enqueued_jobs do
+      assert_broadcasts_text "Test message", to: :messages do |text, target|
+        Message.create(content: "Ignored").broadcast_append_later_to(target, renderable: MessageComponent.new(text))
+      end
+    end
+  end
+  
+  test "Message broadcast prepend later with renderable: render option" do
+    visit messages_path
+    wait_for_stream_to_be_connected
+  
+    perform_enqueued_jobs do
+      assert_broadcasts_text "Test message", to: :messages do |text, target|
+        Message.create(content: "Ignored").broadcast_prepend_later_to(target, renderable: MessageComponent.new(text))
+      end
+    end
+  end
+
   test "Users::Profile broadcasts Turbo Streams" do
     visit users_profiles_path
     wait_for_stream_to_be_connected


app/channels/turbo/streams/broadcasts.rb
```diff
@@ -64,6 +64,11 @@ def broadcast_prepend_later_to(*streamables, **opts)
   end
 
   def broadcast_action_later_to(*streamables, action:, target: nil, targets: nil, **rendering)
+    # Serialize the renderable object if it is present.
+    if rendering[:renderable]
+      rendering[:html] = rendering.delete(:renderable).render_in(self)
+    end
```
seanpdoyle, line 70: Is the property re-assignment necessary? The [`action_view` code](https://github.com/rails/rails/blob/653725ee7cce83d61543183b8305952c0944ec20/actionview/lib/action_view/rendering.rb#L165-L166) that checks for `respond_to?(:render_in)` seems to expect the object under the `:renderable` key.

seanpdoyle: Ah, I see. View Components aren't serializable to Active Job arguments:

```
Error:
BroadcastsTest#test_Message_broadcast_prepend_later_with_renderable:_render_option:
ActiveJob::SerializationError: Unsupported argument type: MessageComponent
    /Users/seanpdoyle/.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/activejob-6.1.7.2/lib/active_job/serializers.rb:30:in `serialize'
    /Users/seanpdoyle/.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/activejob-6.1.7.2/lib/active_job/arguments.rb:122:in `serialize_argument'
```

This means that if we stick with the current proposed change, rendering of the `renderable:` option would occur immediately, **in-process with the request**, and the **broadcast** would be enqueued to a background job.

Prior to this change, the _rendering_ itself occurs in the background job alongside the broadcasting. This change would mean that `rendering:` options are a special case, require special care, and could have adverse effects on the request-response cycle if the rendering is expensive.

I'm curious what it'd take to make ViewComponent `ActiveJob`-serializable.

davidalejandroaguilar: It is.

Since we're rendering the renderable in advance, if we don't re-assign, the `:renderable` key will have a `ActiveSupport::SafeBuffer` value, and `Turbo::Streams::Broadcasts#broadcast_action_to` will call `render_format(:html, **rendering)` causing it to fail with:

```rb
NoMethodError: private method `format' called for "<div class='test-message'>Test message</div>":ActiveSupport::SafeBuffer
```

Because it's not actually a renderable that responds to `format`, but rendered HTML.

davidalejandroaguilar: Yes I was going to mention exactly that: 

That what will happen in the background is the actual broadcasting, but the rendering will still be synchronous. Unless we somehow serialize the component in a way that is compatible with `ActiveJob` / `Sidekiq`

davidalejandroaguilar: >Prior to this change, the rendering itself occurs in the background job alongside the broadcasting. 

Well, not exactly true, prior to this change, these `_later` methods using a renderable are entirely broken 😄

davidalejandroaguilar: I'm thinking we could also try and serialize it ourselves, here are some ideas:

1. Serializing in a custom format that we'll decode later in `Broadcasts#broadcast_action_to` (e.g. checking if `renderable` is a `Hash` with the appropriate values.

```rb
  def broadcast_action_later_to(...)
    if rendering[:renderable]
      rendering[:renderable] = {
        class_name: rendering[:renderable].class.name,
        render_args: # ... some way to get the attributes
      }
    end
# ...
```

2. Using something like `Marshal` `dump`/`load`. Though the same subtleties that exist when serializing `ActiveRecord` instances with this approach exists, like the component class changing or loading unsafe content.

```rb
serialized = Marshal.dump(rendering[:renderable])
"\x04\bo:\x15MessageComponent\x06:\n@textI\"\x11Test message\x06:\x06ET"

Marshal.load(serialized)
#<MessageComponent:0x000000010b8b5798 @text="Test message">

Marshal.load(serialized).render_in(self)
"<div class='test-message'>Test message</div>"
```

Wonder if we should care about this now, or simply fixing the broken `_later` methods is enough for now.

sbounmy:
Following this, having the issue with `_later` and `renderable`.

PR #504, opened 2023-10-16
Support for multiple string arguments to frame helper

### LINKED ISSUE #503: turbo_frame_tag with arrary argument regression
```
<%= turbo_frame_tag [@term_hour.location_id, @term_hour.term.to_param, params[:day_of_week]] do %>
```

With 1.4.0:
```
<turbo-frame id="1_1_sunday">
```

With 1.5.0:

```
<turbo-frame id="1 1 sunday">
```

It seems like #476 broke this behavior.

jcoyne:
This fixes a regression in 1.5.0

Fixes #503

### DIFF
--- a/app/helpers/turbo/frames_helper.rb
+++ b/app/helpers/turbo/frames_helper.rb
@@ -24,6 +24,9 @@ module Turbo::FramesHelper
   #   <% end %>
   #   # => <turbo-frame id="tray"><div>My tray frame!</div></turbo-frame>
   #
+  #   <%= turbo_frame_tag [user_id, "tray"], src: tray_path(tray) %>
+  #   # => <turbo-frame id="1_tray" src="http://example.com/trays/1"></turbo-frame>
+  #
   # The `turbo_frame_tag` helper will convert the arguments it receives to their
   # `dom_id` if applicable to easily generate unique ids for Turbo Frames:
   #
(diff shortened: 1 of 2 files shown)


app/helpers/turbo/frames_helper.rb
```diff
@@ -36,7 +36,7 @@ module Turbo::FramesHelper
   #   <%= turbo_frame_tag(Article.find(1), Comment.new) %>
   #   # => <turbo-frame id="article_1_new_comment"></turbo-frame>
   def turbo_frame_tag(*ids, src: nil, target: nil, **attributes, &block)
```
seanpdoyle, line 38: Could you add an example to the documentation comment above like the one in https://github.com/hotwired/turbo-rails/pull/505/files#diff-7694cf5e712468388f0d32c6638e9d9508b612f0346ce4beb9093960a757fa59R26-R28?

seanpdoyle (approved):

app/helpers/turbo/frames_helper.rb
jcoyne: Yes, I've pulled your example in.

jcoyne:
@seanpdoyle it looks like one of the tests is timing out and I don't have the permissions to restart it.

seanpdoyle:
I don't have permissions either. If you rebase the commit and force push, it'll run again

jcoyne:
@afcapel are you able to restart that one build?

swanson:
Thank you! Ran into this as well when upgrading

PR #511, opened 2023-11-09
Support `targets:` in `Broadcastable` methods - Update

JuannFerrari:
> #210 added the ability to set `targets` on a stream tag. But that doesn't work nicely with the `Broadcastable` helper methods. Currently you have to do this to target some `targets`:
> 
> ```ruby
> after_update_commit -> { broadcast_update_to self, target: nil, targets: ".class_name" }
> ```
> 
> This PR improves things so that you don't need to provide `target: nil` anymore.

--

This PR is a continuation of the work made by @ghiculescu at PR #408


fguillen-getsafe:
Yes, please. It is just weird we have to _desactivate_ `target` to be able to use `targets`. I would even say that the right way is using `target` for both cases, and always using a CSS query selector.

dhh:
If you could rebase, I'd be happy to take this.

JuannFerrari:
@dhh rebase done :)

JamesChevalier:
Is it intended for this to cause `broadcast_remove_to` to require `render: false`?

For example:
`after_destroy -> { broadcast_remove_to("#{user_id}-activities") }`
throws "ActionView::MissingTemplate: Missing partial" errors unless it's rewritten to
`after_destroy -> { broadcast_remove_to("#{user_id}-activities", render: false) }`

Changed in https://github.com/hotwired/turbo-rails/pull/511/commits/5d6e6042d4ff03c0418b7d37e43588f5106e22ec
Later rebased in https://github.com/hotwired/turbo-rails/pull/511/commits/76f492a91085e127a9d0ab16e4be7bf1fce33b6a

seanpdoyle:
@JamesChevalier no, that should not be necessary.

Could you open a new issue to track the resolution of that problem?

PR #514, opened 2023-11-11
Tests: rely on load hooks instead of class constants

seanpdoyle:
Follow-up to [#513][]

Rely on [on_load hooks][] to extend test helpers. Despite the fact that Action Cable is an optional dependency and Action Dispatch is not, this commit replaces all `class`-style extensions with `on_load` hooks for the sake of consistency.

There isn't a hook for `ActionCable::Channel::TestCase`, so waiting for `action_cable_channel` to fire is the best that we can do.

[#513]: https://github.com/hotwired/turbo-rails/pull/513
[on_load hooks]: https://guides.rubyonrails.org/engines.html#available-load-hooks

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -13,14 +13,14 @@ module ActionViewTestCaseExtensions
   end
 end
 
-class ActiveSupport::TestCase
+ActiveSupport.on_load :active_support_test_case do
   include ActiveJob::TestHelper
 end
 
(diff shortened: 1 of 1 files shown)


xymbol:
In #512 and #513, we're dealing with runtime configuration, i.e. apps that load `turbo-rails` but don't require `actioncable`, but this can't happen in tests—unless I'm missing something.

@seanpdoyle In which situation do we need this change?

PR #515, opened 2023-11-11
Add actioncable dependency to gemspec

xymbol:
Turbo depends on ActionCable in runtime, but it's not listed in the gemspec. This is not a problem with the tests because of the Gemfile and, more generally, because ActionCable is a dependency of Rails.

While investigating a loading problem with turbo-rails and an app that didn't use ActionCable, I ran into this.


xymbol:
Since submitting the above, I found this initializer, which might have been the root cause of the `FrozenError` I was looking into. 

https://github.com/hotwired/turbo-rails/blob/e44b6a98a77a0a2cc927f986a67517d73c4c9246/lib/turbo/engine.rb#L19

I'll close this for now.

PR #519, opened 2023-11-23
Require active support module extensions

npezza93:
Without this line, when rails boots thread_mattr_accessor throws an error  saying it's undefined. 

This is happening on 2.0 beta1

### DIFF
--- a/lib/turbo-rails.rb
+++ b/lib/turbo-rails.rb
@@ -1,3 +1,4 @@
+require "active_support/core_ext/module/attribute_accessors_per_thread"
 require "turbo/engine"
 
 module Turbo


seanpdoyle (approved):
@jorgemanrubia is there a way to cover this bug with our current CI environment? Our `test/dummy` application is [hard-coded to depend on Sprockets](https://github.com/hotwired/turbo-rails/blob/4eb4e928e30be8cd537af8073f98b80ddea4a578/test/dummy/config/application.rb#L8).

adrienpoly:
I think it is a duplicate of https://github.com/hotwired/turbo-rails/pull/523 that is already merge on main

PR #521, opened 2023-11-23
Fix sending created broadcasts refreshes to channels where no one listens

tonysm:
### Fixed

- In https://github.com/hotwired/turbo-rails/pull/499 a new `broadcasts_refreshes` method was added that sets up the model hooks to send broadcasts refreshes automatically. However, it repeats the same issue previously found in the `broadcasts` method (https://github.com/hotwired/turbo-rails/pull/295) where it sends the creation broadcasts to the current model's channel and defaults to, but no one will ever be listening on that. This PR allows passing a `stream` param to `broadcasts_refreshes` method, which will be used as the channel name when broadcasting refreshes on model creation and defaults to the model's plural name, similar to how the `broadcasts` method works. It also sends the broadcast refresh right away instead of later when the model is deleted.

### DIFF
--- a/test/dummy/app/models/board.rb
+++ b/test/dummy/app/models/board.rb
@@ -1,5 +1,5 @@
 class Board < ApplicationRecord
-    has_many :tasks
+  has_many :tasks
 
-    broadcasts_refreshes
+  broadcasts_refreshes
 end
--- a/test/dummy/app/models/board.rb
+++ b/test/dummy/app/models/board.rb
@@ -1,0 +1,5 @@
+class Board < ApplicationRecord
+    has_many :tasks
+
+    broadcasts_refreshes
+end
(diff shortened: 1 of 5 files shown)


test/dummy/app/models/board.rb
```diff
@@ -0,0 +1,5 @@
+class Board < ApplicationRecord
+  has_many :tasks
+
+  broadcasts_refreshes
+end
```
northeastprince, line 5: Why is a new model necessary to test this?

tonysm: I added one that actually uses the `broadcasts_refreshes` method. I'm not a Rails dev, so pls let me know if there's another way ;)

northeastprince: I think adding it to an existing model would work just as well.

jorgemanrubia (approved):
Nice @tonysm, thanks for working on this one.

I like the additional model so that we can keep the other models isolated from these additional streams that will get triggered implicitly. I think that isolation is good for testing.

PR #525, opened 2023-11-25
Fix unsupported `ActiveSupport::SafeBuffer` being passed to BroadcastStreamJob

### LINKED ISSUE #522: `Turbo::Streams::BroadcastStreamJob` raising `ActiveJob::SerializationError` with `ActiveSupport::SafeBuffer`
The callbacks defined using `broadcasts_refreshes` (from 2.0.0-beta.1) fail with 

```
Failed enqueuing Turbo::Streams::BroadcastStreamJob to Async(default): ActiveJob::SerializationError (Unsupported argument type: ActiveSupport::SafeBuffer)
```

northeastprince:
Fixes #522.

### DIFF
--- a/app/channels/turbo/streams/broadcasts.rb
+++ b/app/channels/turbo/streams/broadcasts.rb
@@ -70,7 +70,7 @@ module Turbo::Streams::Broadcasts
 
   def broadcast_refresh_later_to(*streamables, request_id: Turbo.current_request_id, **opts)
     refresh_debouncer_for(*streamables, request_id: request_id).debounce do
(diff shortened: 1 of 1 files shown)


seanpdoyle:
I've shared my thoughts in https://github.com/hotwired/turbo-rails/issues/522#issuecomment-1834009762, but to repeat them here:

I wonder if it'd be better to conditionally package an [ActiveJob::Serializers::ObjectSerializer](https://edgeapi.rubyonrails.org/classes/ActiveJob/Serializers/ObjectSerializer.html) to special-case HTML safe strings.

aprescott:
For anyone who might find this in future, this same change was later made in #683 (and included in v2.0.10).

PR #531, opened 2023-11-30
Remove deprecated Turbo::IncludesHelper

brunoprietog:
Since we are updating to the next major version, we could remove the `Turbo::IncludesHelper` deprecated helper.


jorgemanrubia (approved):
Good call @brunoprietog. Ok to merge this one @afcapel?

brunoprietog:
Yeah, tough luck haha. Will it work for a fast 2.0.1 version? Or will we have to wait for 3.0?

PR #534, opened 2023-12-01
Render full documents for requests with `Turbo-Frame:` header

seanpdoyle:
Re-submission of [#232][]
Related to [hotwired/turbo#1047][]

Render full documents, including default layout rendering behavior.

Rendering a minimal layout forces `turbo:reload` events because of the severe difference in the contents of the minimal layout's `<head>` and the requesting document's fully populated `<head>`.

[#232]: https://github.com/hotwired/turbo-rails/pull/232
[hotwired/turbo#1047]: https://github.com/hotwired/turbo/issues/1047

### DIFF
--- a/app/controllers/turbo/frames/frame_request.rb
+++ b/app/controllers/turbo/frames/frame_request.rb
@@ -21,7 +21,6 @@ module Turbo::Frames::FrameRequest
   extend ActiveSupport::Concern
 
   included do
-    layout -> { "turbo_rails/frame" if turbo_frame_request? }
     etag { :frame if turbo_frame_request? }
 
     helper_method :turbo_frame_request?, :turbo_frame_request_id
--- a/app/views/layouts/turbo_rails/frame.html.erb
+++ b/app/views/layouts/turbo_rails/frame.html.erb
@@ -1,8 +1,0 @@
-<html>
-  <head>
-    <%= yield :head %>
-  </head>
-  <body>
-    <%= yield %>
-  </body>
-</html>
--- a/test/frames/frame_request_controller_test.rb
+++ b/test/frames/frame_request_controller_test.rb
@@ -6,7 +6,7 @@ class Turbo::FrameRequestControllerTest < ActionDispatch::IntegrationTest
     assert_select "title", count: 1
 
     get tray_path(id: 1), headers: { "Turbo-Frame" => "true" }
-    assert_select "title", count: 0
+    assert_select "title", count: 1
   end
 
   test "frame request layout includes `head` content" do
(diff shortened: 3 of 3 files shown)


seanpdoyle:
As an alternative, we could send the Turbo Action in the [Turbo-Action:](https://github.com/hotwired/turbo/pull/694) header, then conditionally include the full layout if that header is present.

seanpdoyle:
@jorgemanrubia is this change worth re-considering after the advent of the Page Refresh with Morph mental model (paraphrased as) "render full HTML documents on the server, then let the client decide how to insert that new content into the document".

jorgemanrubia:
@seanpdoyle the removed layout should only kick in for requests with a Turbo Frame header. Could you put an example of how page refreshes with morphing doesn't play well with this? They seem like unrelated systems to me.

seanpdoyle:
> how page refreshes with morphing doesn't play well with this

I'm sorry for being unclear. I'm suggesting that the Turbo Frame server-side mental model change to be **more** like Morphing.


Morphing
---

When submitting a form that redirects back or navigating a link that replaces the page, the server sends a fully formed document. As I understand, the value proposition of a Morphing page refresh is that the server does not care about the diff between what its rendering and what's on the client -- the client will handle negotiating the new content into the existing content. Even if the entirety of the new content is a new flash message, the full `<html>` document is sent down in the request and it's Turbo responsibility to render it. 

Furthermore, the client doesn't communicate anything special to the server. Other than being `fetch`-initiated requests, the server doesn't require any special headers or information in order to integrate with Morphing.

Frames
---

From the client's perspective, frames behave in a similar way. Regardless of the contents of the response HTML, Turbo will find the `<turbo-frame>` element in the response that corresponds to the initiator element based on its `[id]`. Once it's found, it swaps the new children into the document.

From the server's perspective, the similarities break down.  The `Turbo-Frame` header is a client-side implementation detail that servers expect. Instead of rendering a full document without any additional information from the client, servers use the `Turbo-Frame` header to determine the response HTML. In the case of `turbo-rails`, that response HTML can vary wildly based on the presence or absence of that header.

The concept of Frames pre-dates Morphing. In a world where applications are encouraged to enable Morphing for the majority of their applications (instead of other approaches like Frame or Stream form submission responses), any gains won by `turbo-rails`'s server-side decision to skip rendering the page layout are likely to be offset by full document responses elsewhere.

So what?
---

Over time, the `turbo-rails`-driven server-side optimization to return different responses for frames and full pages has led to unintended side-effects and several Issues and codebase workarounds in `@hotwired/turbo`.

The proposal is that by reversing the optimization and re-establishing the traditional Turbolinks expectations (that the servers responds with full HTML documents and the client expects full HTML documents), we have an opportunity to posit a singular simplified mental model and simplify some of `@hotwired/turbo`'s codepaths and patterns.

PR #538, opened 2023-12-07
Fix System Tests in CI environment

seanpdoyle:
Add `rails@7.1` to the CI Test Matrix
===

As part of reading from the `RAILS_VERSION` environment set in CI,
update the `test/dummy/config/application.rb` to use that environment
variable when loading defaults.

Fix a test failure when `rails@7.1` by delegating `Message#to_s` to
`Message#content` so that tests that broadcast messages without text
continue to pass.

Also remove an entry with `ruby: "head"`, since we aren't specifying
`head` in the `ruby:` key.

Fix System Tests in CI environment
===

Resolve breaking changes introduced in `@hotwired/turbo` that were later
fixed in [@hotwired/turbo#4f334da][] (and further resolved in
[@hotwired/turbo#7a7c6e2][]).

CI is failing because those changes are not available as part of a beta
release. Despite the fact that these changes wouldn't be necessary if
that release were available, this resolves the broken build pipeline
more immediately.

[@hotwired/turbo#4f334da]: https://github.com/hotwired/turbo/commit/4f334dab00f0da92c91de652fe7ba2da3a99376b
[@hotwired/turbo#7a7c6e2]: https://github.com/hotwired/turbo/commit/7a7c6e2b83837ff830467106b67f965bca578aa9

Assign `window.Turbo` in `@hotwired/turbo-rails/index`
===

With the changes made in the prior commit to pass CI merged, the fact
that Turbo and Turbo Rails somehow have differing references to `Turbo`
and `window.Turbo`.

This commit reverts the test harness change that sets `window.Turbo`,
and instead assigns that value as part of the
`@hotwired/turbo-rails/index` module:

```diff
-import { Turbo } from "@hotwired/turbo-rails"
-
-window.Turbo = Turbo
+import "@hotwired/turbo-rails"
```


test/dummy/app/javascript/application.js
```diff
@@ -1,2 +1,4 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
-import "@hotwired/turbo-rails"
+import { Turbo } from "@hotwired/turbo-rails"
+
+window.StreamActions = Turbo.StreamActions
```
zzak, line 4: Should this just be `window.Turbo = Turbo`?

seanpdoyle: This change fails `test/system/broadcasts_test.rb`:

```diff
--- a/test/dummy/app/javascript/application.js
+++ b/test/dummy/app/javascript/application.js
@@ -1,4 +1,4 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
-import { Turbo } from "@hotwired/turbo-rails"
+import { Turbo } "@hotwired/turbo-rails"
 
-window.StreamActions = Turbo.StreamActions
+window.Turbo = Turbo
```

MSP-Greg:
@seanpdoyle

Thanks for fixing this.

> Also remove an entry with ruby: "head", since we aren't specifying head in the ruby: key.

Values used in 'include' do not need to be in the normal matrix.  For instance, Ruby Windows has three head builds available in GitHub Actions, `head`, `ucrt`, and `mswin`.  [ruby/openssl](https://github.com/ruby/openssl/blob/1fa9fc54ebc480dcab435212d020ec29068bac48/.github/workflows/test.yml#L18-L31) & [puma/puma](https://github.com/puma/puma/blob/08e71a2c6aaccf295bea909aacf40a9aa86b3702/.github/workflows/tests.yaml#L40-L54) both run `mswin` jobs in their 'include' section.

seanpdoyle:
@jorgemanrubia are you able to review these changes?

jorgemanrubia (approved):
Thank you so much @seanpdoyle

MSP-Greg:
Thanks.  Puma's turbo-rails workflow, which uses both Puma head and turbo-rails head is now passing.
https://github.com/puma/puma/actions/runs/7212424085

Issue #549, opened 2024-01-03
Disable morphing per page


morgoth:
Currently when setting `turbo_refreshes_with(method: :morph, scroll: :preserve)` in layout, it is set globally.

It's not easy to opt out of morphing on some single page. Duplicating meta tags via:
```
<% content_for(:head) do %>
  <%= turbo_refreshes_with(method: :replace, scroll: :reset) %>
<% end %>
```
does not work.

Even such combo doesn't work:
`<%= content_for?(:morphing) ? yield(:morphing) : turbo_refreshes_with(method: :morph, scroll: :preserve) %>`
```
<% content_for(:morphing) do %>
  <%= turbo_refreshes_with(method: :replace, scroll: :reset) %>
<% end %>
```

as the meta tags are cached between page navigation, so they cannot be overwritten this way.

Is there a nice way to do it? Maybe created meta tags should be marked with `"data-turbo-track": "reload"` so it's easier to change them between pages?

PR #550, opened 2024-01-03
Update Turbo Drive `<meta>` across navigations

seanpdoyle:
Closes [#549]

Add System Test level coverage to ensure that Turbo Drive navigations will re-render any `<meta>` elements nested within the document's `<head>`.

To achieve this coverage, introduce the `PagesController#show` action that links to HTML pages that render their `<head>` based on the `turbo_refresh_method` and `turbo_refresh_scroll` query parameters.

[#549]: https://github.com/hotwired/turbo-rails/issues/549

### DIFF
--- a/test/dummy/app/controllers/pages_controller.rb
+++ b/test/dummy/app/controllers/pages_controller.rb
@@ -1,0 +1,2 @@
+class PagesController < ApplicationController
+end
--- a/test/dummy/app/views/pages/show.html.erb
+++ b/test/dummy/app/views/pages/show.html.erb
@@ -1,0 +1,16 @@
+<%= turbo_refreshes_with method: params.fetch(:turbo_refresh_method, :replace).to_sym, scroll: params.fetch(:turbo_refresh_scroll, :reset).to_sym %>
+
+<h1><%= params[:id].titleize %></h1>
+
+<% {
+  classic: {
+    turbo_refresh_method: :replace,
+    turbo_refresh_scroll: :reset,
+  },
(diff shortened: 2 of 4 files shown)


morgoth:
@seanpdoyle Thank you for looking into this.

So I dig a bit more into it.
The behavior varies on tag placement.
```
<%= yield(:head) %>
<%= turbo_refreshes_with(method: :morph, scroll: :preserve) %>
```
does not work at all on first render. This is reasonable as the `turbo_refreshes_with` tries to write into head that was already yielded.

With layout:
```
<%= turbo_refreshes_with(method: :morph, scroll: :preserve) %>
<%= yield(:head) %>
```
And different page:
```
<%= turbo_refreshes_with(method: :replace, scroll: :reset) %>
```
I'm getting doubled tags when navigating from root page to the given page.

```
<meta name="turbo-refresh-method" content="morph">
<meta name="turbo-refresh-scroll" content="preserve">
...other head content...
<meta name="turbo-refresh-method" content="replace">
<meta name="turbo-refresh-scroll" content="reset">
```
and this seems to work properly as the behavior are taken from last tags on the page.

However when I enter the page directly (or refresh in the browser), I'm getting tags in such order:
```
<meta name="turbo-refresh-method" content="replace">
<meta name="turbo-refresh-scroll" content="reset">
<meta name="turbo-refresh-method" content="morph">
<meta name="turbo-refresh-scroll" content="preserve">
```
thus, it does not work as expected as the morphing is the last one.

PR #553, opened 2024-01-11
Test installer

zzak:
This commit adds test coverage for the installer Rake task and application template. The installer is run against a freshly generated Rails app using the version of Rails that is currently loaded. Thus the installer can be tested with different versions of Rails in CI.

Similar to hotwired/stimulus-rails#136

The motivation is to be able to move these tests from railties, see: https://github.com/rails/rails/pull/49679


zzak:
FWIW, this morning I also looked at adding a test for [turbo_needs_redis](https://github.com/hotwired/turbo-rails/blob/main/lib/install/turbo_needs_redis.rb), but couldn't find an easy way to mock whether or not redis was available on the system.

This method here determines that, but it's not easily stubbed since the tasks are run in a different process:
https://github.com/hotwired/turbo-rails/blob/38a7ca6467385eaea2efb5ee037dd0fad4a10572/lib/tasks/turbo_tasks.rake#L8-L12

I still think this could happen after, if anyone comes up with a clever way to test it, and we've at least increased the test coverage here so that we can continue on the railties side.

zzak:
@afcapel Sorry for the ping, would you be able to review this one and hotwired/stimulus-rails#136?

zzak:
@afcapel Thanks for your review! I've addressed your comments. :bow:

I think Rails 6.1 is broken, and this test will always fail:

```
Error:
InstallerTest#test_installer:
Errno::ENOENT: No such file or directory @ rb_sysopen - app/javascript/application.js
    test/installer_test.rb:15:in `read'
    test/installer_test.rb:15:in `block (2 levels) in <class:InstallerTest>'
    test/installer_test.rb:8:in `block in <class:InstallerTest>'
```

This is because the check always assumes that `app/javascript/application.js` exists, but in Rails 6.1 it's using webpack, see: 5c5a85da8aafe3a7af919a8b8e05daafd53fab60

```
/tmp/d20240419-2363909-lm6npn/turbo_test_app_ba923836b46f828536c9eea7e80c44f44a9ee2cf => cat app/javascript/packs/application.js
// This file is automatically compiled by Webpack, along with any other files
// present in this directory. You're encouraged to place your actual application logic in
// a relevant structure within app/javascript and only use these pack files to reference
// that code so it'll be compiled.

import Rails from "@rails/ujs"
import Turbolinks from "turbolinks"
import * as ActiveStorage from "@rails/activestorage"
import "channels"

Rails.start()
Turbolinks.start()
ActiveStorage.start()
```

Unless I'm missing something? :thinking: I thought this test was passing before, but if you think of anything obvious LMK!

zzak:
:facepalm: I've fixed the tests for Rails 6.1, the red text made me thing it was an error but the install does complete and `yarn add ...` still works, so I guess we're good!

zzak:
@afcapel Could I trouble you for another review? :bow:

Issue #554, opened 2024-01-11
Suggestion: turbo_stream.clear


svoop:
The following code – shamelessly borrowed from the [tutorial on Hotrails](https://www.hotrails.dev/turbo-rails/turbo-frames-and-turbo-streams) – is needed in order to empty a Turbo frame:

```ruby
<%= turbo_stream.update Quote.new, "" %>
```

Would you consider a PR which adds the following doing exactly the same?

```ruby
<%= turbo_stream.clear Quote.new %>
```

seanpdoyle:
I've opened https://github.com/hotwired/turbo-rails/pull/555 to document how applications might extend the existing methods provided by the `turbo_stream` helper.

svoop:
@seanpdoyle Thanks for the extra docs. I take, `.clear` is not as common a case as I initially thought (sort of as a companion to `.remove`) and therefore doesn't merit to be implemented in Turbo itself. You know this much better than me, so a custom helper it is then.

seanpdoyle:
@svoop to clarify, #555 is more than a documentation change, it also adds support. In the meantime, you can call `turbo_stream.action`:

```ruby
turbo_stream.action :clear, Quote.new
```

alexandreruban:
@svoop I proposed this change earlier in https://github.com/hotwired/turbo-rails/pull/356 but it wasn't accepted.

You can have a look at the code and implement it yourself in your project thanks to the PR @seanpdoyle mentioned.

svoop:
@alexandreruban Ah, `.empty`, I searched for clear (Array-like) and clean, sorry for the dupe. It sure is great to have the option to add it on a project level now.

link0ff:
But still please add `turbo_stream.refresh`.  Indeed, it's possible now to add it as a custom action.  And also it's possible to use `turbo_stream.turbo_stream_action_tag(:refresh)`. 

However, it's so useful to employ this action instead of doing redirect that doesn't do morphing, and moreover `refresh` already belongs to the set of default actions, so that it would be nice to have such a shorthand as just `turbo_stream.refresh`.

marcoroth:
I've been building Turbo Power to account for more common stream actions, which also exposes a `turbo_frame_reload` action:

https://github.com/marcoroth/turbo_power

Feel free to use/copy out the specific stream actions you need for your project.

PR #555, opened 2024-01-11
Document how to extend `turbo_stream` for custom actions

seanpdoyle:
Closes [#554][]

The current `Turbo::Streams::TagBuilder` class supports the actions that Turbo provides out of the box.

For applications that define [Custom Stream Actions][], this new support aims to provide them with a mechanism to define Rails-side helpers.

For example, consider an `[action="highlight"]` custom stream. To build those with first-party tooling, applications could add a [hook block][] to their initializer:

  ```ruby
  # config/initializers/turbo.rb
  ActiveSupport.on_load :turbo_streams_tag_builder do
    def highlight(target)
      action :highlight, target
    end

    def highlight_all(targets)
      action_all :highlight, targets
    end
  end

  turbo_stream.highlight "my-element"
  # => <turbo-stream action="highlight" target="my-element"><template></template></turbo-stream>

  turbo_stream.highlight_all ".my-selector"
  # => <turbo-stream action="highlight" targets=".my-selector"><template></template></turbo-stream>
  ```

[#554]: https://github.com/hotwired/turbo-rails/issues/554
[Custom Stream Actions]: https://turbo.hotwired.dev/handbook/streams#custom-actions
[hook block]: https://edgeapi.rubyonrails.org/classes/ActiveSupport/LazyLoadHooks.html#method-i-on_load

### DIFF
--- a/app/models/turbo/streams/tag_builder.rb
+++ b/app/models/turbo/streams/tag_builder.rb
@@ -22,6 +22,24 @@
 #   <%= turbo_stream.append dom_id(topic_merge) do %>
 #     <%= link_to topic_merge.topic.name, topic_path(topic_merge.topic) %>
 #   <% end %>
+#
+# To integrate with custom actions, extend this class in response to the :turbo_streams_tag_builder load hook:
+#
+#   ActiveSupport.on_load :turbo_streams_tag_builder do
+#     def highlight(target)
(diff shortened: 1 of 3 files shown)


afcapel:
Thanks @seanpdoyle!

PR #557, opened 2024-01-13
Render `<turbo-cable-stream-source>` as `display: none`

seanpdoyle:
Since the `<turbo-cable-stream-source>` is a Custom Element, it has a default `style:` property of `display: inline`. As a visible element, it participates in the page's layout.

This commit changes the element to always be `display: none` so that it no longer participates in the layout.


jordan-brough:
Here's an example of `<turbo-cable-stream-source>` messing up the layout:

This code:
```html
<style>
.grid-container {
  display: grid;
  grid-template-columns: repeat(2, 1fr);
}
.grid-item {
  background-color: lightblue;
}
</style>

<div class="grid-container">
  <div class="grid-item">Item 1</div>
  <div class="grid-item">Item 2</div>
  <turbo-cable-stream-source></turbo-cable-stream-source>
  <div class="grid-item">Item 3</div>
  <div class="grid-item">Item 4</div>
</div>
```

produces this result:
image

See [this jsfiddle](https://jsfiddle.net/ou5emgvt/1/).

It feels like this shouldn't be the default behavior for `<turbo-cable-stream-source>` elements.

jordan-brough:
We ran into this also. I've opened a github issue for the topic here: https://github.com/hotwired/turbo-rails/issues/643

Would be great to see this or some other solution implemented.

PR #561, opened 2024-01-22
Show actual assert count next to expected count

matteeyah:
## Summary

When assert against a specific count of turbo streams, the error message can be confusing. When you assert 2 streams, but there is only one, the error message says there's no streams - (`Expected 2 Turbo Stream broadcasts on "Z2lkOi8vb3B0b25hbC9SZWNvcmRpbmcvMjcyMzgxNjI3:analysis", but there were none`). This is confusing, because it implies that here were no streams, but in fact, it reports this error message when the expected count is not the same as the asserted count.

## Proposal

Show the actual count next to the expected count when asserting against a specific count.

### DIFF
--- a/lib/turbo/broadcastable/test_helper.rb
+++ b/lib/turbo/broadcastable/test_helper.rb
@@ -64,7 +64,7 @@ module Turbo
         else
           broadcasts = "Turbo Stream broadcast".pluralize(count)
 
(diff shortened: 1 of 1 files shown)


seanpdoyle (approved):

matteeyah:
@afcapel Can you take a look at this simple change, please?

afcapel:
Thanks @matteeyah

PR #562, opened 2024-01-23
Fix double registration of :turbo_stream Renderer

skipkayhil:
When an application loads both `ActionController::Base` and
`ActionController::API`, the `:action_controller` load hook gets called
twice (once for each of them). Since the `:turbo_stream` Renderer is
registered in this load hook, it ends up getting registered twice and
prints a warning for method redefinition.

To demonstrate the warning in the `turbo-rails` tests, an
`Api::ApplicationController` was added to the dummy app so that both
base controller classes get loaded:

```
$ CI=true bundle exec rake test
/home/hartley/.cache/asdf/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/actionpack-7.1.3/lib/action_controller/metal/renderers.rb:75: warning: method redefined; discarding old _render_with_renderer_turbo_stream
/home/hartley/src/github.com/skipkayhil/turbo-rails/lib/turbo/engine.rb:67: warning: previous definition of _render_with_renderer_turbo_stream was here
```

This commit fixes the issue by not using the `:action_controller` load
hook to register the `:turbo_stream` Renderer. Adding a Renderer should
not require a lazy load hook because `ActionController::Renderers` does
not load any other classes.

### DIFF
--- a/test/dummy/app/mailboxes/application_mailbox.rb
+++ b/test/dummy/app/mailboxes/application_mailbox.rb
@@ -1,2 +1,0 @@
-class ApplicationMailbox < ActionMailbox::Base
-end
--- a/lib/turbo/engine.rb
+++ b/lib/turbo/engine.rb
@@ -63,11 +63,9 @@ module Turbo
     end
 
     initializer "turbo.renderer" do
-      ActiveSupport.on_load(:action_controller) do
-        ActionController::Renderers.add :turbo_stream do |turbo_streams_html, options|
-          self.content_type = Mime[:turbo_stream] if media_type.nil?
-          turbo_streams_html
-        end
+      ActionController::Renderers.add :turbo_stream do |turbo_streams_html, options|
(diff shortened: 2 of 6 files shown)


seanpdoyle (approved):

test/dummy/app/mailboxes/application_mailbox.rb
seanpdoyle, line 1: Is the removal of the Mailboxes related to engine changes?

skipkayhil: It was to enable `eager_load` since neither the Action Mailbox nor the Action Mailer railties are required in `config/application.rb`. I think `eager_load` has merits on its own, but it also makes it easier to demonstrate the issue in this PR because we don't have to _use_ the `Api::ApplicationController` in a test.

brunoprietog:
Thanks @skipkayhil!

PR #565, opened 2024-02-02
Include broadcastable assertions automatically in ActiveSupport::TestCase

swanson:
There is a note in the Campfire codebase `test_helper`:

image

And I noticed I had the same patch in my own codebase:

image

In https://github.com/hotwired/turbo-rails/pull/466 it explicitly mentioned the `Broadcastable::TestHelpers` are "not automatically included" -- but I'm not sure why. @seanpdoyle any insights?

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ The [`Turbo::TestAssertions`](./lib/turbo/test_assertions.rb) concern provides T
 
 The [`Turbo::TestAssertions::IntegrationTestAssertions`](./lib/turbo/test_assertions/integration_test_assertions.rb) are built on top of `Turbo::TestAssertions`, and add support for passing a `status:` keyword. They are automatically included in [`ActionDispatch::IntegrationTest`](https://edgeguides.rubyonrails.org/testing.html#integration-testing).
 
-The [`Turbo::Broadcastable::TestHelper`](./lib/turbo/broadcastable/test_helper.rb) concern provides Action Cable-aware test helpers that assert that `<turbo-stream>` elements were or were not broadcast over Action Cable. They are not automatically included. To use them in your tests, make sure to `include Turbo::Broadcastable::TestHelper`.
+The [`Turbo::Broadcastable::TestHelper`](./lib/turbo/broadcastable/test_helper.rb) concern provides Action Cable-aware test helpers that assert that `<turbo-stream>` elements were or were not broadcast over Action Cable. `Turbo::Broadcastable::TestHelper` is automatically included in [`ActiveSupport::TestCase`](https://edgeapi.rubyonrails.org/classes/ActiveSupport/TestCase.html).
 
 ## Development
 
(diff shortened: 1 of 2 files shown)


swanson:
I could also see `Turbo::TestAssertions` including `Turbo::Broadcastable::TestHelpers` as well so we just include one module in the railtie if that would be preferable.

seanpdoyle:
I don't remember specifically, but in retrospect, I think the transitive `ActionCable::TestHelper` made me hesitant.

Since that isn't included globally, it seemed appropriate to follow suit with these helpers too.

afcapel:
Thanks @swanson 

I think we can include this change as part of a minor release, as it won't cause any harm to people that were already including the module themselves.

```ruby
module Greetings
  def greet
    "Hello!"
  end
end

class Person
  include Greetings
  include Greetings # Including the module a second time
end

person = Person.new
puts person.greet # Outputs: "Hello!"
```

edariedl:
@swanson @afcapel Hi, we are using `turbo-rails` gem in our Rails app but we are not using the `actioncable`. 

With Rails 7.1 upgrade we had to start to including `ActionCable` into the app, because `turbo-rails` failed during initialization without it in production environment. This change is now causing the following error in our tests probably because of a missing configuration (because we don't need one for our app).
```
Error:
InvoicePaymentTest#test_pay_foreign_invoice:
NoMethodError: undefined method `fetch' for nil
    /Users/edadriedl/.rbenv/versions/3.3.0-rc1/lib/ruby/gems/3.3.0+0/gems/actioncable-7.1.3/lib/action_cable/server/configuration.rb:38:in `pubsub_adapter'
    /Users/edadriedl/.rbenv/versions/3.3.0-rc1/lib/ruby/gems/3.3.0+0/gems/actioncable-7.1.3/lib/action_cable/server/base.rb:86:in `block in pubsub'
    /Users/edadriedl/.rbenv/versions/3.3.0-rc1/lib/ruby/gems/3.3.0+0/gems/actioncable-7.1.3/lib/action_cable/server/base.rb:86:in `synchronize'
    /Users/edadriedl/.rbenv/versions/3.3.0-rc1/lib/ruby/gems/3.3.0+0/gems/actioncable-7.1.3/lib/action_cable/server/base.rb:86:in `pubsub'
    /Users/edadriedl/.rbenv/versions/3.3.0-rc1/lib/ruby/gems/3.3.0+0/gems/actioncable-7.1.3/lib/action_cable/test_helper.rb:10:in `before_setup'
    /Users/edadriedl/.rbenv/versions/3.3.0-rc1/lib/ruby/gems/3.3.0+0/gems/activerecord-7.1.3/lib/active_record/test_fixtures.rb:11:in `before_setup'
    /Users/edadriedl/.rbenv/versions/3.3.0-rc1/lib/ruby/gems/3.3.0+0/gems/activesupport-7.1.3/lib/active_support/testing/setup_and_teardown.rb:40:in `before_setup'
```

Is there a way to fix it without adding the complete infratructure for action cable into our app? Thank you.

PR #566, opened 2024-02-06
Use content_tag instead of tag helper

pranavbabu:
This PR enhances performance by replacing the usage of `tag` with `tag_builder.tag_string`. 
This change eliminates the overhead of going through method_missing, resulting in faster HTML generation.

### DIFF
--- a/app/helpers/turbo/frames_helper.rb
+++ b/app/helpers/turbo/frames_helper.rb
@@ -42,6 +42,6 @@ module Turbo::FramesHelper
     id = ids.first.respond_to?(:to_key) ? ActionView::RecordIdentifier.dom_id(*ids) : ids.join('_')
     src = url_for(src) if src.present?
 
-    tag_builder.tag_string(:turbo_frame, **attributes.merge(id: id, src: src, target: target).compact, &block)
+    content_tag(:turbo_frame, nil, **attributes.merge(id: id, src: src, target: target).compact, &block)
   end
 end


app/helpers/turbo/frames_helper.rb
```diff
@@ -42,6 +42,6 @@ def turbo_frame_tag(*ids, src: nil, target: nil, **attributes, &block)
     id = ids.first.respond_to?(:to_key) ? ActionView::RecordIdentifier.dom_id(*ids) : ids.join('_')
     src = url_for(src) if src.present?
 
-    tag.turbo_frame(**attributes.merge(id: id, src: src, target: target).compact, &block)
+    tag_builder.tag_string(:turbo_frame, **attributes.merge(id: id, src: src, target: target).compact, &block)
```
seanpdoyle, line 45: The [ActionView::Helpers::TagHelper::TagBuilder](https://github.com/rails/rails/blob/6b2a9f1213d7aa4f9a25db7640ad68ec0d8c31ce/actionview/lib/action_view/helpers/tag_helper.rb#L47) is marked as `:nodoc:`, so it isn't intended to be part of the public API.

If performance is a concern, is there an opportunity to replace `tag.turbo_frame` with [content_tag](https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-content_tag):

```suggestion
    content_tag(:turbo_frame, nil, **attributes.merge(id: id, src: src, target: target).compact, &block)
```

seanpdoyle, line 45: The [ActionView::Helpers::TagHelper::TagBuilder](https://github.com/rails/rails/blob/6b2a9f1213d7aa4f9a25db7640ad68ec0d8c31ce/actionview/lib/action_view/helpers/tag_helper.rb#L47) is marked as `:nodoc:`, so it isn't intended to be part of the public API.

If performance is a concern, is there an opportunity to replace `tag.turbo_frame` with [content_tag](https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-content_tag):

```suggestion
    content_tag(:turbo_frame, nil, **attributes.merge(id: id, src: src, target: target).compact, &block)
```

seanpdoyle (changes requested):

app/helpers/turbo/frames_helper.rb
```diff
@@ -42,6 +42,6 @@ def turbo_frame_tag(*ids, src: nil, target: nil, **attributes, &block)
     id = ids.first.respond_to?(:to_key) ? ActionView::RecordIdentifier.dom_id(*ids) : ids.join('_')
     src = url_for(src) if src.present?
 
-    tag.turbo_frame(**attributes.merge(id: id, src: src, target: target).compact, &block)
+    content_tag(:turbo_frame, nil, **attributes.merge(id: id, src: src, target: target).compact, &block)
```
brunoprietog, line 45: According to the [content_tag Rails API documentation](https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-content_tag), the use of `content_tag` is considered legacy syntax.

brunoprietog (commented):
Sorry, but I think this shouldn't be done. Using tag builder as it is now is the right way [according to the Rails API documentation](https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-tag)

seanpdoyle:
While I appreciate the performance concerns, @brunoprietog is correct: the `tag` builder will be the way forward. Optimizations and configuration support to extend the set of supported tags are likely to land in Action View's future.

As I understand it, the difference between rendering a known element through a defined method and an unknown element through method missing is in the range of millisecond (worst case) and microseconds (best case).

Are your applications rendering enough `<turbo-frame>` elements in their templates to experience significant performance degradation?

Issue #573, opened 2024-02-09
uninitialized constant Turbo::Broadcastable::TestHelper::ActionCable


jcoyne:
When generating a Rails app with `--skip-action-cable` and then installing turbo rails 2.0.2 I see this error when I load the environment:

```
NameError:
  uninitialized constant Turbo::Broadcastable::TestHelper::ActionCable
```

This was not a problem with turbo-rails 1.5.0

jcoyne:
@swanson could this have been introduced by https://github.com/hotwired/turbo-rails/pull/565 ?

taylorthurlow:
Looks like you beat me to this one by 10 minutes. Yeah it's definitely the test helper assuming that ActionCable is available. I believe the entire test helper should be able to be skipped if ActionCable isn't present, the helper isn't useful otherwise.

seanpdoyle:
I've opened https://github.com/hotwired/turbo-rails/pull/574 to try to address this.

I'm a little hazing on the loading order quirks, so could you pull it down and test it out with your application's test harnesses?

PR #574, opened 2024-02-09
Only include `Turbo::Broadcastable::TestHelper`  with Action Cable

seanpdoyle:
Closes [#573][]
Related to [#565][]

Re-structure the automatic inclusion of the
`Turbo::Broadcastable::TestHelper` module so that it's only automatically loaded and installed when Action Cable is available to the runtime.

[#573]: https://github.com/hotwired/turbo-rails/issues/573
[#565]: https://github.com/hotwired/turbo-rails/pull/565#issuecomment-1935720027

### DIFF
--- a/lib/turbo/engine.rb
+++ b/lib/turbo/engine.rb
@@ -81,10 +81,14 @@ module Turbo
     initializer "turbo.test_assertions" do
       ActiveSupport.on_load(:active_support_test_case) do
         require "turbo/test_assertions"
-        require "turbo/broadcastable/test_helper"
-
         include Turbo::TestAssertions
-        include Turbo::Broadcastable::TestHelper
+      end
+
+      ActiveSupport.on_load(:action_cable) do
+        ActiveSupport.on_load(:active_support_test_case) do
+          require "turbo/broadcastable/test_helper"
+          include Turbo::Broadcastable::TestHelper
+        end
       end
 
       ActiveSupport.on_load(:action_dispatch_integration_test) do


lib/turbo/engine.rb
```diff
@@ -81,10 +81,13 @@ class Engine < Rails::Engine
...
         require "turbo/test_assertions"
-        require "turbo/broadcastable/test_helper"
 
         include Turbo::TestAssertions
-        include Turbo::Broadcastable::TestHelper
+
+        if defined? ActionCable
+          require "turbo/broadcastable/test_helper"
+          include Turbo::Broadcastable::TestHelper
+        end
```
afcapel, line 90: @seanpdoyle what do you think about adding a `ActiveSupport.on_load(:action_cable)` block here? The Rails guides [suggest this technique to execute code that depends on several dependencies](https://edgeapi.rubyonrails.org/classes/ActiveSupport/LazyLoadHooks.html).

t27duck:
Cautiously optimistic, but this branch seems to fix the test suite for my personal app.

lib/turbo/engine.rb
seanpdoyle: I like that idea. I was unsure about nesting, but I'll try it out.

afcapel:
Thanks @seanpdoyle 🙏

joshuay03:
@seanpdoyle Is this planned to be released / back-ported anytime soon?

nflorentin:
@seanpdoyle 

Upgrading from 2.0.0 to 2.0.2 breaks with the following error :
```
uninitialized constant Turbo::Broadcastable::TestHelper::ActionCable (NameError)

          include ActionCable::TestHelper
```

I use turbo-rails without ActionCable, I don't need broadcasting.

I don't understand why Turbo::Broadcastable::TestHelper is included, it seems to make turbo-rails not usable without action cable.

seanpdoyle:
@nflorentin this change is not included in `2.0.2`. Does depending on `main` resolve the issue?

aaronjensen:
Asking a question here with the best intentions given that this change fixes a breaking change that prevents people from updating to 2.X. Is there a significant impediment in the release process that prevents an immediate release after a fix like this is merged in order to unblock people that are downstream? If so, is there any way that I could help alleviate it? In a few of the Rails-adjacent projects (propshaft, turbo, turbo-rails) I've noticed that things get merged (which is great) but then releases don't happen for sometimes weeks or months. I know everyone is busy doing their full-time job so I'm not asking for heroics, I'm genuinely curious if there's anything I can do to help alleviate the cost of a release? Thank you!

seanpdoyle:
@afcapel @jorgemanrubia could either of you consider a fast-tracked 2.0.3 release?

afcapel:
I've just published 2.0.3 👍

aaronjensen:
Thank you!

PR #577, opened 2024-02-11
Introduce `Turbo::SystemTestHelper`

seanpdoyle:
Introduce the `Turbo::SystemTestHelper` module to be included into
[ActionDispatch::SystemTestCase][] when it's available.

The module is named to mimic [ActionText::SystemTestHelper][].

The module defines a `#connect_turbo_cable_stream_sources` helper
method extracted from this project's System Test suite. It aims to
synchronize the test harness with Turbo's Action Cable-powered broadcast
support. The method will find all `<turbo-cable-stream-source>` elements
that are present but not yet `[connected]` (returning the results
immediately with Capybara's `:wait`), then wait for them to connect
(using whatever Capybara's configured wait value).

In addition to the `connect_turbo_cable_stream_sources`, also introduce
a `:turbo_cable_stream_source` Capybara selector, along with
`assert_turbo_cable_stream_source` and
`assert_no_turbo_cable_stream_source` helper methods.

[ActionDispatch::SystemTestCase]: https://edgeapi.rubyonrails.org/classes/ActionDispatch/SystemTestCase.html
[ActionText::SystemTestHelper]: https://edgeapi.rubyonrails.org/classes/ActionText/SystemTestHelper.html

### DIFF
--- a/lib/turbo/engine.rb
+++ b/lib/turbo/engine.rb
@@ -5,6 +5,7 @@ module Turbo
     isolate_namespace Turbo
     config.eager_load_namespaces << Turbo
     config.turbo = ActiveSupport::OrderedOptions.new
+    config.turbo.test_connect_after_actions = %i[visit]
     config.autoload_once_paths = %W(
       #{root}/app/channels
       #{root}/app/controllers
@@ -148,6 +149,25 @@ module Turbo
           end
 
           @encoders[:turbo_stream] = TurboStreamEncoder.new
+        end
+      end
+    end
+
+    initializer "turbo.system_test_helper" do
+      ActiveSupport.on_load(:action_dispatch_system_test_case) do
+        require "turbo/system_test_helper"
+        include Turbo::SystemTestHelper
+      end
+    end
+
+    config.after_initialize do |app|
+      ActiveSupport.on_load(:action_dispatch_system_test_case) do
+        app.config.turbo.test_connect_after_actions.map do |method|
+          class_eval <<~RUBY, __FILE__, __LINE__ + 1
+            def #{method}(...)                                    # def visit(...)
+              super.tap { connect_turbo_cable_stream_sources }    #   super.tap { connect_turbo_cable_stream_sources }
+            end                                                   # end
+          RUBY
         end
       end
     end
--- a/lib/turbo/system_test_helper.rb
+++ b/lib/turbo/system_test_helper.rb
@@ -1,0 +1,128 @@
+module Turbo::SystemTestHelper
+  # Delay until every `<turbo-cable-stream-source>` element present in the page
+  # is ready to receive broadcasts
+  #
+  #   test "renders broadcasted Messages" do
+  #     message = Message.new content: "Hello, from Action Cable"
+  #
+  #     visit "/"
+  #     click_link "All Messages"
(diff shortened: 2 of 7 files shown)


lib/turbo/system_test_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+  #
+  #   test "Page receives new Messages broadcast over Action Cable" do
+  #     visit messages_path
+  #     wait_for_turbo_cable_stream_sources
+  #
+  #     # create a new Message, then broadcast...
+  #
+  #     assert_text "Hello from Action Cable"
+  #   end
+  def wait_for_turbo_cable_stream_sources(**options, &block)
```
seanpdoyle, line 15: I don't love this name, as the name itself is composed mostly of implementation details. To be honest, it's annoying and tedious to have to wait at all.

What would be _really_ nice (ergonomically) would be built-in support for Capybara's actions like `visit`, `click_link`, `click_button`, etc. We _could_ override those methods like:

```ruby
def visit(...)
  super
  wait_for_turbo_cable_stream_sources
end
```

That feels pretty invasive and presumptive, but it papers over the tedium.

Regardless, documenting the call is something we'll need to do regardless. The most important part of documentation is the name itself, so we'll want to come up with something that's an improvement on `wait_for_turbo_cable_stream_sources`.

seanpdoyle: I've pushed up a change to extend `#visit`. Does that feel like a good balance between surprise, risk, and convenience?

seanpdoyle:
@jorgemanrubia @afcapel have you encountered this in your test suites?

lib/turbo/engine.rb
```diff
@@ -5,6 +5,7 @@ class Engine < Rails::Engine
     isolate_namespace Turbo
     config.eager_load_namespaces << Turbo
     config.turbo = ActiveSupport::OrderedOptions.new
+    config.turbo.waiting_capybara_actions = %i[visit]
```
seanpdoyle, line 8: This name needs to be improved upon.

lib/turbo/system_test_helper.rb
```diff
@@ -0,0 +1,35 @@
...
+  #
+  #   # config/environments/test.rb
+  #   config.turbo.waiting_capybara_actions << :click_link
+  #
+  # To disable automatic waiting, set the configuration to `[]`:
+  #
+  #   # config/environments/test.rb
+  #   config.turbo.waiting_capybara_actions = []
+  #
+  def wait_for_turbo_cable_stream_sources(**options)
```
seanpdoyle, line 30: As an alternative to "wait", does "connect" feel better?

```ruby
def connect_turbo_cable_stream_sources(**options)
```

seanpdoyle:
@brunoprietog are you able to review this?

seanpdoyle:
@jorgemanrubia are you able to review this proposal?

seanpdoyle:
@brunoprietog are you able to review this?

gjtorikian:
Bless you and your bumps across PRs @seanpdoyle — I see you and thank you.

jorgemanrubia (approved):
Great one @seanpdoyle

Issue #579, opened 2024-02-13
`refresh` action is missing from Turbo::Streams::TagBuilder


phil-6:
Attempting to user `format.turbo_stream { render turbo_stream: turbo_stream.refresh('target') }`

Raises a `NoMethodError`: `undefined method `refresh' for an instance of Turbo::Streams::TagBuilder`

I think the new `refresh` action needs to be added into the [TagBuilder](https://github.com/hotwired/turbo-rails/blob/07d3488c12774b59c6c8eb94836019557d5e2160/app/models/turbo/streams/tag_builder.rb#L43) model

seanpdoyle:
@jorgemanrubia is the `<turbo-stream action="refresh" request-id="...">` element meant for public consumption, or is it an implementation detail of `broadcasts_refreshes`?

seanpdoyle:
@nikolaokonesh your question does not seem related to this issue, and might be better suited as an Issue of its own.

With that being said, are you confident that the `<meta>` elements are being rendered into the document's `<head>`? If you call `turbo_refreshes_with` **after** the `yield :head` call, does it behave as you'd expect?

nikolaokonesh:
@seanpdoyle  I tried other options, it doesn't work.

seanpdoyle:
I've opened https://github.com/hotwired/turbo-rails/pull/595 to implement this.

PR #595, opened 2024-03-03
Add `turbo_stream.refresh` builder method

seanpdoyle:
Closes [#579][]

Extends the `turbo_stream` tag builder helper to create `<turbo-stream action="refresh">` elements through the pre-existing `turbo_stream_refresh_tag` method.

[#579]: https://github.com/hotwired/turbo-rails/issues/579

### DIFF
--- a/app/models/turbo/streams/tag_builder.rb
+++ b/app/models/turbo/streams/tag_builder.rb
@@ -228,6 +228,18 @@ class Turbo::Streams::TagBuilder
     action_all :prepend, targets, content, **rendering, &block
   end
 
+  # Creates a `turbo-stream` tag with an `[action="refresh"`] attribute and a
+  # `[request-id]` attribute that defaults to `Turbo.current_request_id`:
+  #
+  #   turbo_stream.refresh
+  #   # => <turbo-stream action="refresh" request-id="ef083d55-7516-41b1-ad28-16f553399c6a"></turbo-stream>
+  #
+  #   turbo_stream.refresh request_id: "abc123"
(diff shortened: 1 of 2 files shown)


OlegChuev:
Hi guys! Any updates/plans/ETA for this PR to be merged?

radanskoric:
Would be nice to have this merged.

rubiii:
Anything we can help to get this shipped?

seanpdoyle:
@jorgemanrubia what do you think about expanding the Tag Builder interface to support `[action="refresh"]`?

adrianthedev:
I was just looking for this method.
Not having it looks like the API is incomplete.
It forces devs to create `turbo_stream.erb` files instead of handling it in the controller.

radanskoric:
> I was just looking for this method. Not having it looks like the API is incomplete. It forces devs to create `turbo_stream.erb` files instead of handling it in the controller.

@adrianthedev, while I hope this gets merged soon, you can still do it in the controller:
```
respond_to do |format|
  format.turbo_stream { render turbo_stream: turbo_stream.action(:refresh, "") }
end
```
This is also useful if you have your own custom actions.

radanskoric (approved):

brendon:
This would be great to have, though in my experience, adding the `request-id` means that the browser will ignore the refresh request (just like it ignores broadcast refresh requests with a matching `request-id`), or does this only apply to streamed refresh requests?

seanpdoyle:
@jorgemanrubia @brunoprietog are either of you available to review this change?

Issue #587, opened 2024-02-17
rails tests fail after turbo-rails upgrade from 1.5.0 to 2.0.3


xeron:
```
NoMethodError: undefined method `fetch' for nil:NilClass
    vendor/bundle/ruby/3.1.0/gems/actioncable-7.1.3/lib/action_cable/server/configuration.rb:38:in `pubsub_adapter'
    vendor/bundle/ruby/3.1.0/gems/actioncable-7.1.3/lib/action_cable/server/base.rb:86:in `block in pubsub'
    vendor/bundle/ruby/3.1.0/gems/actioncable-7.1.3/lib/action_cable/server/base.rb:86:in `synchronize'
    vendor/bundle/ruby/3.1.0/gems/actioncable-7.1.3/lib/action_cable/server/base.rb:86:in `pubsub'
    vendor/bundle/ruby/3.1.0/gems/actioncable-7.1.3/lib/action_cable/test_helper.rb:10:in `before_setup'
    vendor/bundle/ruby/3.1.0/gems/activerecord-7.1.3/lib/active_record/test_fixtures.rb:11:in `before_setup'
    vendor/bundle/ruby/3.1.0/gems/activesupport-7.1.3/lib/active_support/testing/setup_and_teardown.rb:40:in `before_setup'
```

morgoth:
Most likely there is no `config/cable.yml` file in your app.

xeron:
> Most likely there is no `config/cable.yml` file in your app.

You're right. I don't use actioncable tho and the app itself works just fine without this file.

t27duck:
If you have `require rails/all` in your `application.rb` file or have the action cable railtie required, then action cable is loaded up. The test helpers that were added in https://github.com/hotwired/turbo-rails/commit/ef4a624de2be9b5241fe97e6c286b4bfd992fcde fires up code... somewhere, that makes action cable want to set up a server in tests which looks for `config/cable.yml`. Missing the file results in that exception.

Manfred:
If Action Cable is now required to use Turbo in Rails, it could be good to add it to `turbo-rails.gemspec`.

AxelTheGerman:
Also noticed this after upgrading. I don't have `require "action_cable/engine"` in my app nor `require "rails/all"`. My tests pass but when preloading the application it fails with:

```
[...]

PR #601, opened 2024-03-10
Revise how to disable eager loading app/channels

### LINKED ISSUE #512: Can't load turbo-rails without Action Cable starting in Rails 7.1.2

fxn:
(This is the same as #513, which I closed by accident.)

Fixes #512.

### DIFF
--- a/lib/turbo/engine.rb
+++ b/lib/turbo/engine.rb
@@ -15,8 +15,25 @@ module Turbo
       #{root}/app/jobs
     )
 
+    # If the parent application does not use Action Cable, app/channels cannot
+    # be eager loaded, because it references the ActionCable constant.
+    #
+    # When turbo-rails depends on Rails 7 or above, the entire block can be
+    # reduced to
+    #
+    #   unless defined?(ActionCable)
+    #     Rails.autoloaders.once.do_not_eager_load("#{root}/app/channels")
+    #   end
+    #
     initializer "turbo.no_action_cable", before: :set_eager_load_paths do
-      config.eager_load_paths.delete("#{root}/app/channels") unless defined?(ActionCable)
+      unless defined?(ActionCable)
+        if Rails.autoloaders.zeitwerk_enabled?
+          Rails.autoloaders.once.do_not_eager_load("#{root}/app/channels")
+        else
+          # This else branch only runs in Rails 6.x + classic mode.
+          config.eager_load_paths.delete("#{root}/app/channels")
+        end
+      end
     end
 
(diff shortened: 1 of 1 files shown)


dorianmariecom:
Thanks @dhh <3

dorianmariecom:
@dhh Can you release a new version? That would be awesome

t27duck:
Until a new version is released, this is the workaround I've been using that seems to work out well enough...

`config/application.rb`:

```ruby
require_relative "boot"

require "rails"
# Pick the frameworks you want:
require "active_model/railtie"
...
# require "action_cable/engine"
...

# Require the gems listed in Gemfile, including any gems
# you've limited to :test, :development, or :production.
Bundler.require(*Rails.groups)

# Hack as turbo-rails loads the constant from the eager load paths.
# Remove once a new version is released.
class Turbo::StreamsChannel; end unless defined?(ActionCable)
```

bruno-:
@afcapel any chance you can release a new gem version with this fix? 🙏

This is currently breaking my app in production (I have to resort to workarounds) because I'm using [mission_control-jobs](https://github.com/rails/mission_control-jobs) which is using turbo.

Many thanks

AxelTheGerman:
Would also appreciate a new release that includes this fix. That should also solve #587 - I was just looking into fixing that and then found the code that already fixed it is there and merged but not released.

Thank you!

PR #602, opened 2024-03-11
Do not load ActiveJobs if ActiveJob is not loaded

aaronjensen:
Fixes https://github.com/hotwired/turbo-rails/issues/478

Replaces https://github.com/hotwired/turbo-rails/pull/479

Based on https://github.com/hotwired/turbo-rails/pull/601

### DIFF
--- a/lib/turbo/engine.rb
+++ b/lib/turbo/engine.rb
@@ -91,8 +91,10 @@ module Turbo
     end
 
     initializer "turbo.broadcastable" do
-      ActiveSupport.on_load(:active_record) do
-        include Turbo::Broadcastable
+      ActiveSupport.on_load(:active_job) do
+        ActiveSupport.on_load(:active_record) do
+          include Turbo::Broadcastable
+        end
       end
     end
 
@@ -120,10 +122,12 @@ module Turbo
         include Turbo::TestAssertions
       end
 
-      ActiveSupport.on_load(:action_cable) do
-        ActiveSupport.on_load(:active_support_test_case) do
-          require "turbo/broadcastable/test_helper"
-          include Turbo::Broadcastable::TestHelper
+      ActiveSupport.on_load(:active_job) do
+        ActiveSupport.on_load(:action_cable) do
+          ActiveSupport.on_load(:active_support_test_case) do
+            require "turbo/broadcastable/test_helper"
+            include Turbo::Broadcastable::TestHelper
+          end
         end
       end
 
--- a/lib/turbo/engine.rb
+++ b/lib/turbo/engine.rb
@@ -15,6 +15,27 @@ module Turbo
       #{root}/app/jobs
     )
 
+    # If the parent application does not use Active Job, app/jobs cannot
+    # be eager loaded, because it references the ActiveJob constant.
+    #
+    # When turbo-rails depends on Rails 7 or above, the entire block can be
+    # reduced to
+    #
+    #   unless defined?(ActiveJob)
+    #     Rails.autoloaders.once.do_not_eager_load("#{root}/app/jobs")
+    #   end
+    #
+    initializer "turbo.no_active_job", before: :set_eager_load_paths do
+      unless defined?(ActiveJob)
+        if Rails.autoloaders.zeitwerk_enabled?
+          Rails.autoloaders.once.do_not_eager_load("#{root}/app/jobs")
+        else
+          # This else branch only runs in Rails 6.x + classic mode.
+          config.eager_load_paths.delete("#{root}/app/jobs")
+        end
+      end
+    end
+
     # If the parent application does not use Action Cable, app/channels cannot
     # be eager loaded, because it references the ActionCable constant.
     #
--- a/test/broadcastable/test_helper_test.rb
+++ b/test/broadcastable/test_helper_test.rb
@@ -1,263 +1,269 @@
 require "test_helper"
 
-class Turbo::Broadcastable::TestHelper::CaptureTurboStreamBroadcastsTest < ActiveSupport::TestCase
-  include Turbo::Broadcastable::TestHelper
+module Turbo
+  module Broadcastable
+    module TestHelper
+      class CaptureTurboStreamBroadcastsTest < ActiveSupport::TestCase
+        include Turbo::Broadcastable::TestHelper
 
-  test "#capture_turbo_stream_broadcasts returns <turbo-stream> elements broadcast on a stream name" do
-    message = Message.new(id: 1)
+        test "#capture_turbo_stream_broadcasts returns <turbo-stream> elements broadcast on a stream name" do
+          message = Message.new(id: 1)
 
-    message.broadcast_replace_to "messages"
-    message.broadcast_remove_to "messages"
-    replace, remove, *rest = capture_turbo_stream_broadcasts "messages"
+          message.broadcast_replace_to "messages"
+          message.broadcast_remove_to "messages"
+          replace, remove, *rest = capture_turbo_stream_broadcasts "messages"
 
-    assert_empty rest
-    assert_equal "replace", replace["action"]
-    assert_equal "remove", remove["action"]
-    assert_not_empty replace.at("template").element_children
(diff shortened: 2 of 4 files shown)


lib/turbo/engine.rb
```diff
@@ -15,6 +15,27 @@ class Engine < Rails::Engine
...
+    #
+    # When turbo-rails depends on Rails 7 or above, the entire block can be
+    # reduced to
+    #
+    #   unless defined?(ActiveJob)
+    #     Rails.autoloaders.once.do_not_eager_load("#{root}/app/jobs")
+    #   end
+    #
+    initializer "turbo.no_active_job", before: :set_eager_load_paths do
+      unless defined?(ActiveJob)
```
zzak, line 29: Would this work instead?

`ActiveSupport.on_load :active_job do`

aaronjensen: I don't know. I'm following the pattern used in the Action Cable check below.

zzak: Correct, sorry I was thinking of the opposite case but this initializer occurs in the absence of AJ :pray:

NakagawaMakoto:
I really do want this PR to be merged.

I have created my project with the following;
  rails new myproj --skip-action-mailer --skip-action-mailbox --skip-action-text --skip-active-record --skip-active-job --skip-active-storage --skip-action-cable --skip-jbuilder --skip-system-test --skip-rubocop --skip-ci --skip-devcontainer --skip-git --skip-docker --css sass --asset-pipeline=propshaft

It works fine in development/test mode, but fails to start with the following in production mode with eager_load set to true;
  c:/ruby33-x64/lib/ruby/gems/3.3.0/gems/turbo-rails-2.0.6/app/jobs/turbo/streams/action_broadcast_job.rb:2:in `<main>': uninitialized constant ActiveJob (NameError)

I do not really think my project gain much by setting eager_load to true, but it would be better not to be bothered by ActiveJob when you know you explicitly exclude the feature.

dhh:
I can't see how this is going to work. Then all the "broadcast_later" calls are just going to fail at runtime. I appreciate wanting a minimal version of Rails for whatever reasons, but this turbo-rails integration is premised on the full integration with broadcasts and what not. I think you're probably better off just pulling in Turbo as a JavaScript dependency and forgoing this integration gem all together.,

aaronjensen:
@dhh It'd work because someone who wasn't using ActiveJob also wouldn't be using `broadcast_later`. The gem itself is useful to people that aren't because of the stream related helpers. Ultimately, this is a cohesion problem. One could, as you said, use the JavaScript dependency directly and implement their own turbo helpers (or copy the relevant ones from this package).

The idea with this was to make this package work for anyone to use the way they wanted to use it. Rails itself is modular enough to allow opting out of ActiveJob. I thought it made sense for other packages that are bundled with Rails to respect the modularity decisions of the users. If someone ends up with a runtime error, they end up with a runtime error. They'd discover that well before they went to production and add the active_job dependency. 

Ideally the "later" versions of methods wouldn't be available at all, so the `broadcast_later` macro would fail at eval time.

dhh:
Do explore it further. But I can't take something where methods are added that then just don't work. It has to be cohesive, as you say. What's there, what's added, is going to work.

aaronjensen:
Would you be amenable to the Broadcastable concern not being mixed in to the model at all if ActiveJob is unavailable?

dhh:
Something like that, but it has to be cohesive. Across the line, we can't have code that's dead at runtime, if we allow this to work without activejob. And the complexity of introducing that has to be proportionate to the fact that this is a minority concern and that the vast majority of users will not do this.

aaronjensen:
@dhh see the updated PR. I exclude Broadcastable and its associated test helper. I also added a note to the documentation within Broadcastable. As far as I can tell, this is cohesive. If you opt out of ActiveJob, you won't get Broadcastable or any of its associated jobs. If you include ActiveJob, you will.

aaronjensen:
One moment, fixing what seems like a local test running issue.

test/broadcastable/test_helper_test.rb
```diff
@@ -1,263 +1,269 @@
...
-      # no-op
-    end
-
-    assert_empty streams
-  end
-end
+module Turbo
+  module Broadcastable
+    module TestHelper
+      class CaptureTurboStreamBroadcastsTest < ActiveSupport::TestCase
```
aaronjensen, line 6: This is necessary because `Turbo::Broadcastable::TestHelper` isn't loaded and therefore the namespace isn't available until after the `include Turbo::Broadcastable::TestHelper`, due to lazy loading. We could use `if defined?(ActiveJob)`, rather than `ActiveSupport.on_load(:active_job)`, but that doesn't seem better.

aaronjensen:
Fixed. One caveat: https://github.com/hotwired/turbo-rails/pull/602/files?w=1#r1760143885

dhh:
That works 👍

aaronjensen:
Great, thank you!

botandrose:
@aaronjensen @dhh I think there's still more to do here. Bumping from v2.0.6. to v2.0.8 broke my app because `Broadcastable` is no longer included, even though I have `require "rails/all"` in my config/application.rb. Trying to narrow down to a failing test.

aaronjensen:
@botandrose in the interim could you post the stack trace? I've tried a few things myself and haven't seen any failures.

botandrose:
Hi @aaronjensen, thanks for the quick response.

My stack trace is just a method missing error for one the broadcast methods, so its not particularly helpful, but I have just confirmed its a regression, and not something weird I'm doing in my app:

Here is fresh Rails 7.2.1 app with zero changes from `rails new`, and you can see that `ApplicationRecord` is missing `Turbo::Broadcastable`. Downgrading to turbo-rails from 2.0.8 to 2.0.7 brings the module back:

```bash
$ rvm use ruby-3.3.4@turbo-rails
Using /home/micah/.rvm/gems/ruby-3.3.4 with gemset turbo-rails

$ rails new bug-turbo-rails && cd bug-turbo-rails
<snip>

$ bin/rails runner 'puts ApplicationRecord.included_modules.include?(Turbo::Broadcastable)'
false

$ vi Gemfile # pin turbo-rails to 2.0.7

$ bundle update turbo-rails
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Resolving dependencies...
Fetching turbo-rails 2.0.7 (was 2.0.8)
Installing turbo-rails 2.0.7 (was 2.0.8)
Note: turbo-rails version regressed from 2.0.8 to 2.0.7
Bundle updated!

$ bin/rails runner 'puts ApplicationRecord.included_modules.include?(Turbo::Broadcastable)'
true
```

I'm not sure why the tests aren't catching this, but it seems clear that its a regression for the common case.

aaronjensen:
That's not good. I'm sorry about that. Looking now.

aaronjensen:
@dhh pinpointed, PR available: https://github.com/hotwired/turbo-rails/pull/678

aaronjensen:
@botandrose if you'd be able change the `initializer "turbo.broadcastable"` in your locally installed gems to the below and test, that would be appreciated.

```ruby
    initializer "turbo.broadcastable" do
      ActiveSupport.on_load(:active_record) do
        if defined?(ActiveJob)
          include Turbo::Broadcastable
        end
      end
    end
```

botandrose:
@aaronjensen Confirmed that this change fixes the issue in both the fresh rails app and in my specific app!

aaronjensen:
@botandrose Thank you for confirming. @dhh https://github.com/hotwired/turbo-rails/pull/678 should be ready to merge and release. I'm standing by if anything else needs to be adjusted.

dhh:
@aaronjensen Thanks! I have a few comments on that PR.

dhh:
Released fix in https://github.com/hotwired/turbo-rails/releases/tag/v2.0.9

botandrose:
Thanks for the quick turnaround, guys!

PR #603, opened 2024-03-14
Only pass `:request_id` option to Refresh Stream

seanpdoyle:
When rendering `<turbo-stream>` elements for broadcasting, **omit** the `:request_id` partial-local variable if it's blank.

If rendered partials utilize [Action View strict locals][], the extra variable raises an error:

```
unknown local :request_id
```

By omitting the `:request_id` from the
`Turbo::Broadcastable#broadcast_rendering_with_defaults`, it isn't passed as part of partial rendering.

Since `broadcast_refresh` never renders a partial itself (it builds an empty `<turbo-stream action="refresh"
request-id="..."></turbo-stream>` element), the `:request_id` local variable isn't necessary.

[Action view strict locals]: https://guides.rubyonrails.org/action_view_overview.html#strict-locals

### DIFF
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -508,7 +508,7 @@ module Turbo::Broadcastable
       options.tap do |o|
         # Add the current instance into the locals with the element name (which is the un-namespaced name)
         # as the key. This parallels how the ActionView::ObjectRenderer would create a local variable.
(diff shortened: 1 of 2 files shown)


earnold:
I just ran into this same issue. Would love to see this PR merged. Is there anyway I can be helpful with it?

seanpdoyle:
@jorgemanrubia are you able to review this change?

jorgemanrubia:
Thanks @seanpdoyle

PR #611, opened 2024-04-10
Fix method references in documentation.

Schwad:
There is a `broadcast_refresh` method and a `broadcasts_refreshes` method, but not a `broadcast_refreshes` method.

I found this out the hard way when following the documentation today!


Schwad:
Yes. I'll close this. Thank you. I don't know if the other method doc correction in that PR is right but it does include my change

Issue #613, opened 2024-04-11
turbo_frame_request_id safe operator in not handled correctly


jwoodrow:
When rendering a layout from specs, the request value will be `nil` but using the safe operator followed by the `[]` method leads to an `undefined method [] for nil`

We render an `axlsx` in some of our specs to validate the content of the rendered file like this

```ruby
ApplicationController.new.render_to_string(
  formats:  [:xlsx],
  handlers: [:axlsx],
  filename:,
  template: 'path/to/view',
  locals:
)
```

this ends up failing here https://github.com/hotwired/turbo-rails/blob/102a491754d46f7dd924201fcfaf879a0f04b11c/app/controllers/turbo/frames/frame_request.rb#L36

```ruby
def turbo_frame_request_id
  request&.headers["Turbo-Frame"]
end
```

I would have used `request&.headers&.dig('Turbo-Frame')` to avoid this.

We found this out because we tried setting up `rails/mission_control-jobs` which depends on `turbo-rails` (which we don't currently use)

seanpdoyle:
@jwoodrow thank you for opening this issue. Could you share a script that reproduces the behavior you're experiencing? I've shared a [reproduction script](https://github.com/hotwired/turbo-rails/pull/593) below that you could modify to suit your needs. For example, you could replace the `SystemTest` and the HTML beneath the `__END__` separator with an `ActiveSupport::TestCase` test that's closer to the source of your issue.

Based on what you've shared above, I'm unsure how the `#[]` access is raising a NoMethodError. the `&` operator should prevent the rest of the expression from executing when `request == nil`. That means the `request` object is present, but the `#headers` is returning `nil`. I'm not sure what could be the root cause of those circumstances.

<details>
<summary>Save this snippet to `bug.rb` then execute `ruby bug.rb`</summary>

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  gem "propshaft"
  gem "puma"
  gem "sqlite3"
  gem "turbo-rails"

  gem "capybara"
  gem "cuprite", require: "capybara/cuprite"
end

ENV["DATABASE_URL"] = "sqlite3::memory:"
ENV["RAILS_ENV"] = "test"

require "active_record/railtie"
# require "active_storage/engine"
require "action_controller/railtie"
require "action_view/railtie"
# require "action_mailer/railtie"
# require "active_job/railtie"
require "action_cable/engine"
# require "action_mailbox/engine"
[...]

Issue #529, opened 2023-11-28
Issue with render_to_string method of ActionController. "Could not render layout: undefined method `headers`"


limjinsun:
```
Ruby version 3.2.2
Rails version 7.0.8
```
GemFile
```
source 'https://rubygems.org'

ruby '3.2.2'

gem 'rails', '7.0.8'
gem 'puma', '~> 4.3.9'
gem 'sass-rails', '~>6.0.0'
gem 'jquery-rails'
gem 'sprockets', '3.7.2'

# gem 'bootstrap', '~> 5.1.3'

gem 'sprockets-rails'
gem 'turbo-rails'
gem 'stimulus-rails'
gem 'hotwire-rails'
gem 'importmap-rails'
gem 'jsbundling-rails'

gem 'jbuilder', '~> 2.5'
gem 'twilio-ruby', '5.72.1'
gem 'google-authenticator-rails'
gem 'friendly_id', '~> 5.2.0'
gem 'mysql2'

...
```

Error.
```
irb(main):003:0> ActionController::Base.new.render_to_string( template: "shared/file_template/test2" )
/Users/xxx/Documents/projects/rails-project/vendor/cache/ruby/3.2.0/gems/turbo-rails-1.5.0/app/controllers/turbo/frames/frame_request.rb:36:in `turbo_frame_request_id': Could not render layout: undefined method `headers' for nil:NilClass (NoMethodError)

      request.headers["Turbo-Frame"]
             ^^^^^^^^
/Users/xxx/Documents/projects/rails-project/vendor/cache/ruby/3.2.0/gems/turbo-rails-1.5.0/app/controllers/turbo/frames/frame_request.rb:36:in `turbo_frame_request_id': undefined method `headers' for nil:NilClass (NoMethodError)

      request.headers["Turbo-Frame"]
             ^^^^^^^^
```

brunoprietog:
I've opened #532 to solve this

ryanckulp:
i have the same problem. perhaps `&.` isn't the best solution, but if there's a way to pass in a `request` object to `render_to_string` then we're all set.

ryanckulp:
hey @brunoprietog , a simple solution i found is to simply pass in an empty Request object:

```rb
ac = ActionController::Base.new
ac.request = ActionDispatch::Request.new({})
ac.render_to_string(template: 'view/partial', locals: { variable: 'here' }, layout: 'application')

# => renders HTML string
```

PR #532, opened 2023-11-30
Use safe operator when calling to request in turbo_frame_request_id

brunoprietog:
Fixes #529

PR #616, opened 2024-04-15
Improve rendering outside request

seanpdoyle:
Follow-up to [#529][]
Closes [#613][]

Remove the `&` operator from the `Turbo::Frames::FrameRequest` concern,
since it isn't solving the original issues in the way that it intended.

Instead, add documentation to the `README.md` that highlights
`turbo-rails`'s compatibility with [ActionController::Renderer][].

[#529]: https://github.com/hotwired/turbo-rails/issues/529
[#613]: https://github.com/hotwired/turbo-rails/issues/613
[ActionController::Renderer]: https://api.rubyonrails.org/classes/ActionController/Renderer.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ When the user clicks on the `Edit this todo` link, as a direct response to this 
 
 ### A note on custom layouts
 
-In order to render turbo frame requests without the application layout, Turbo registers a custom [layout method](https://api.rubyonrails.org/classes/ActionView/Layouts/ClassMethods.html#method-i-layout). 
+In order to render turbo frame requests without the application layout, Turbo registers a custom [layout method](https://api.rubyonrails.org/classes/ActionView/Layouts/ClassMethods.html#method-i-layout).
 If your application uses custom layout resolution, you have to make sure to return `"turbo_rails/frame"` (or `false` for TurboRails < 1.4.0) for turbo frame requests:
 
 ```ruby
@@ -64,7 +64,7 @@ layout :custom_layout
 
 def custom_layout
   return "turbo_rails/frame" if turbo_frame_request?
-  
+
   # ... your custom layout logic
 ```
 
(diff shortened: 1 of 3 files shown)


seanpdoyle:
@jwoodrow thank you for sharing more clarifying details to https://github.com/hotwired/turbo-rails/issues/613.

I've opened this PR to add stronger guarding conditionals instead of using the `&` safe access operator. 

Having fixed the issue, I want to acknowledge that I dislike adding these kinds of conditionals to concerns that are mixed into a controller context, since the presence of a request is crucial to that context and feels like a safe assumption to have been making.

In #613 you've shared some sample code:

```ruby
ApplicationController.new.render_to_string(
  formats:  [:xlsx],
  handlers: [:axlsx],
  filename:,
  template: 'path/to/view',
  locals:
)
```

If this call were split up into two successive calls, you'd have an opportunity to assign a request instance:

```ruby
controller = ApplicationController.new
controller.request = ActionDispatch::Request.new({})

controller.render_to_string(
  formats:  [:xlsx],
  handlers: [:axlsx],
  filename:,
  template: 'path/to/view',
  locals:
)
```

If you made that change to your tests, would they pass?

Could you share more context about your particular use case? I wonder if there's a different solution to the underlying problem.

eugeneius:
Rails provides `ActionController:Renderer` for rendering templates outside of a request:
https://api.rubyonrails.org/v7.1.3.2/classes/ActionController/Renderer.html

Internally it builds a dummy request object:
https://github.com/rails/rails/blob/v7.1.3.2/actionpack/lib/action_controller/renderer.rb#L124

The sample code from https://github.com/hotwired/turbo-rails/issues/613 should be rewritten as:

```ruby
ApplicationController.renderer.render_to_string(
  formats:  [:xlsx],
  handlers: [:axlsx],
  filename:,
  template: 'path/to/view',
  locals:
)
```

I think instead of this change, https://github.com/hotwired/turbo-rails/pull/532 should be reverted, since it didn't fix anything and was attempting to support usage that isn't Rails public API.

seanpdoyle:
@eugeneius great point. I've updated this PR to be more focused on documentation instead.

PR #618, opened 2024-04-16
Adds external_link_to helper to navigate outside of a turbo-frame

Schwad:
SEEKING FEEDBACK:

I feel that it would be a clearer API for newer users to have an explicit `external_link_to` method instead of defining `link_to "..", target: "_top"` every time they want to link outside of a turbo-frame.

If this is accepted by maintainers I will add tests and can tweak the documentation if needed.

### DIFF
--- a/app/helpers/turbo/link_to_helper.rb
+++ b/app/helpers/turbo/link_to_helper.rb
@@ -1,0 +1,13 @@
+module Turbo::LinkToHelper
+  # Allows creation of a link that can navigate outside of the turbo frame:
+  #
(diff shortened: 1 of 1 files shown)


meneerprins:
In my mind `external_link_to` would mean a `link_to` with `target: '_blank'` since it links to an 'external' page, also isn't it supposed to be `data-turbo-frame="_top"` instead of `target="_top"`?

brunoprietog:
Hi @Schwad, thanks for your contribution! I also agree with @meneerprins, so I'll close this

PR #641, opened 2024-06-29
README: Using local Turbo version

klevo:
While contributing to [Turbo](https://github.com/hotwired/turbo) it took me a while to figure out how to package my local version with turbo-rails. I also noticed in some other discussions people struggling with this. Documenting it here for future reference.


brunoprietog:
Thanks @klevo!

PR #647, opened 2024-07-16
Remove morph broadcast helpers

omarluq:
Follow up to morph action API restructure by @seanpdoyle in https://github.com/hotwired/turbo/pull/1240
This PR Removes morph stream action broadcast helpers and updates unit test and documentation for morph usage

### DIFF
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -260,6 +260,10 @@ module Turbo::Broadcastable
   #   # Sends <turbo-stream action="replace" target="clearance_5"><template><div id="clearance_5">Other partial</div></template></turbo-stream>
   #   # to the stream named "identity:2:clearances"
   #   clearance.broadcast_replace_to examiner.identity, :clearances, partial: "clearances/other_partial", locals: { a: 1 }
+  #
+  #   # Sends <turbo-stream action="replace" method="morph" target="clearance_5"><template><div id="clearance_5">Other partial</div></template></turbo-stream>
+  #   # to the stream named "identity:2:clearances"
+  #   clearance.broadcast_replace_to examiner.identity, :clearance, attributes: { method: :morph }, partial: "clearances/other_partial", locals: { a: 1 }
   def broadcast_replace_to(*streamables, **rendering)
     Turbo::StreamsChannel.broadcast_replace_to(*streamables, target: self, **broadcast_rendering_with_defaults(rendering)) unless suppressed_turbo_broadcasts?
   end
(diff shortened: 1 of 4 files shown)


app/models/concerns/turbo/broadcastable.rb
```diff
@@ -279,6 +283,10 @@ def broadcast_replace(**rendering)
   #   # Sends <turbo-stream action="update" target="clearance_5"><template><div id="clearance_5">Other partial</div></template></turbo-stream>
   #   # to the stream named "identity:2:clearances"
   #   clearance.broadcast_update_to examiner.identity, :clearances, partial: "clearances/other_partial", locals: { a: 1 }
+  #
+  #   # sends <turbo-stream action="update" method="morph" target="clearance_5"><template><div id="clearance_5">Other partial</div></template></turbo-stream>
```
seanpdoyle, line 287: Where does the `[method="morph"]` come from in these code samples?

Would the default value of the `:method` attribute be `"replace"`?

test/streams/streams_channel_test.rb
```diff
@@ -286,31 +286,53 @@ class Turbo::StreamsChannelTest < ActionCable::Channel::TestCase
     end
   end
 
-  test "broadcasting morph now" do
-    options = { partial: "messages/message", locals: { message: "hello!" } }
-  
-    assert_broadcast_on "stream", turbo_stream_action_tag("morph", target: "message_1", template: render(options)) do
-      Turbo::StreamsChannel.broadcast_morph_to "stream", target: "message_1", **options
+  test "broadcasting actions with method morph now" do
+    options = { method: :morph, partial: "messages/message", locals: { message: "hello!" } }
```
seanpdoyle, line 290: When does `:method` need to be nested within an `:attributes` Hash option, and when can it be passed at the top level?

omarluq: This was a miss on my part it should always be an optional key under attributes.

app/models/concerns/turbo/broadcastable.rb
omarluq: It's an optional attribute, updated the docs to reflect it.

omarluq: Yes the default value would be replace but we don't need to explicitly pass it as you know! in case the method attribute is not present on the element the action will be processed as replace right?

jorgemanrubia (approved):
Thank you so much @omarluq 🙏

searls:
I just tried implementing this after the release of 2.0.6 and wanted to share a challenge I had.

The first thing I tried was adding `morph:` to a `turbo_stream.update` method:

```erb
<%= turbo_stream.update @dom_id, method: :morph do %>
…
```

This didn't change the rendering of the tag, so I tried:

```erb
<%  turbo_stream_action_tag :update, target: @dom_id, method: :morph do %>
…
```

Which truncated the content in the block (whether or not I wrapped it in `<template>`).

This was the only formulation that worked:

```erb
<%= tag.turbo_stream(action: :update, target: @dom_id, method: :morph) do %>
  <template>…</template>
<% end %>
```

Finally, after reading this PR thread, I learned about the `attributes` keyword option:

```erb
<%= turbo_stream.update @dom_id, attributes: {method: :morph} do %>
  …
<% end %>
```

Which seems to be the best working way to use the helper with the new morph update feature.

Path of least surprise: does it seem like the convenience methods of `turbo_stream.update` and `turbo_stream.replace` should support a top-level `method` param?

seanpdoyle:
@searls from an ergonomics perspective, I think that special treatment of a `:method` keyword argument would be valuable. Having said that, for every method-level keyword option that we support, that's one fewer local variables the view partials will have available to them through their `local_assigns`. 

If a view template relied on local assignments like `:attributes, `:target`, and `:method`, could they be surprisingly `nil` or fail strict local assertions?

searls:
@seanpdoyle I'd leave up whether to pull up `method` as a special case up to you all, but I guess what I'm running into is just a sense of inconsistency. For example, I just had to update the above from:

```erb
<%= turbo_stream.update @dom_id, attributes: {method: :morph} do %>
```

To `update_all`, as the replaced item could appear multiple times on the page:

```erb
<%= turbo_stream.update_all @selector, attributes: {method: :morph} do %>
```

And it immediately stopped using morphing. 

Because it's (apparently?) hard to tell if morphing is happening (I had to drop a debugger inside Turbo's source), asymmetries like this in the API could lead to people accidentally quietly disabling the feature without knowing it.

brendon:
It's not super clear, but was this ever resolved? I'm struggling to find any way to render an `update` with `method: 'morph'` from a controller. The `turbo_stream.update` method doesn't seem to allow any `attributes` passthrough to `turbo_stream_action_tag`.

PR #656, opened 2024-08-01
ci: Automate requesting review

omarluq:
adds a simple github action that automates requesting review from @jorgemanrubia and @seanpdoyle on PR's that passes CI

### DIFF
--- a/.github/workflows/request-review.yaml
+++ b/.github/workflows/request-review.yaml
@@ -1,0 +1,24 @@
+name: Request Review
+on:
+  workflow_run:
+    workflows: ["CI"]
+    types:
+      - completed
+
+jobs:
+  request_review:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - name: Request reviews
+        uses: actions/github-script@v6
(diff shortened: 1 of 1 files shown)


.github/workflows/request-review.yaml
```diff
@@ -0,0 +1,24 @@
...
+  request_review:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - name: Request reviews
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const reviewers = ['seanpdoyle', 'jorgemanrubia'];
```
seanpdoyle, line 18: I'm flattered to be included in this list.

With that being said, I am not a project maintainer, and do not have permissions to merge or close pull requests.

omarluq: your opinions and feedback are valuable imho and defiantly should be on the list. 
That being said hardcoding usernames on github actions is bad practice, this pr is to stir up a convo about this but ideally there should be a github team that the maintainers controller membership to and gets tagged instead

PR #658, opened 2024-08-04
Add attributes hash as optional arg to replace/update model tag builders

omarluq:


jorgemanrubia (changes requested):
@omarluq instead of allowing arbitrary attributes I'd rather add a dedicated `method:` key. Could you please change that?

Thanks for working on this.

omarluq:
> instead of allowing arbitrary attributes I'd rather add a dedicated `method:` key

@jorgemanrubia I will open a new PR that adds `method:` as top lvl optional arg for both broadcast tag builders and model tag builders. 🫡

AlexKovynev:
> > instead of allowing arbitrary attributes I'd rather add a dedicated `method:` key
> 
> @jorgemanrubia I will open a new PR that adds `method:` as top lvl optional arg for both broadcast tag builders and model tag builders. 🫡

@omarluq and... ? :)

tonywok:
For those of you looking for a work-around in the meantime:

```
turbo_stream.action(:replace, "your-id", attributes: { method: "morph" }) do
  # Your content
end
```

hjhart:
Tracking this just now as I'm trying to use morphing in a rails project. Does this give us a good reason to release a new version of `turbo-rails`?

PR #657, opened 2024-08-02
Fix deprecation in routing

atomaka:
Fixes new deprecation in Rails from https://github.com/rails/rails/pull/52422 that marks paths as hash keys in routes for removal.

PR #660, opened 2024-08-06
Fix deprecated routes

siaw23:
Drawing a route with a hash key name is deprecated and will be removed in Rails 8.1.

Reference: https://github.com/rails/rails/pull/52422


brunoprietog:
Hi @siaw23, thanks for your contribution, but this is a duplicate of #657, so closing this one

PR #661, opened 2024-08-07
Raise ArgumentError when turbo_stream_from helper gets passed blank streamables

mwallba:
Hi everyone,

When people pass in blank streamables into `turbo_stream_from` it can lead to unpredictable issues for i.e. the postgresql ActionCable adapter blowing up (postgresql's NOTIFY will throw an exception when passed in an empty string):
https://github.com/reclaim-the-stack/actioncable-enhanced-postgresql-adapter/issues/3 (this issue does also happen with the regular postgresql adapter - I ran into it myself).

I'm not sure if it is a little too overzealous here to prevent any of the passed in streamables from being blank. But I can't come up with a scenario where one might pass in a blank string or nil in between other streamables intentionally.

### DIFF
--- a/app/helpers/turbo/streams_helper.rb
+++ b/app/helpers/turbo/streams_helper.rb
@@ -49,12 +49,12 @@ module Turbo::StreamsHelper
   #
   #   <%= turbo_stream_from "room", channel: RoomChannel, data: {room_name: "room #1"} %>
   #
-  # Raises an +ArgumentError+ if any streamables are blank
+  # Raises an +ArgumentError+ if all streamables are blank
   #
   #   <%= turbo_stream_from("") %> # => ArgumentError: streamables can't be blank
-  #   <%= turbo_stream_from("channel", nil, "messages") %> # => ArgumentError: streamables can't be blank
+  #   <%= turbo_stream_from("", nil) %> # => ArgumentError: streamables can't be blank
   def turbo_stream_from(*streamables, **attributes)
-    raise ArgumentError, "streamables can't be blank" if streamables.any?(&:blank?)
+    raise ArgumentError, "streamables can't be blank" unless streamables.any?(&:present?)
     attributes[:channel] = attributes[:channel]&.to_s || "Turbo::StreamsChannel"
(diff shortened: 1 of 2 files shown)


app/helpers/turbo/streams_helper.rb
```diff
@@ -48,7 +48,13 @@ def turbo_stream
   # It is also possible to pass additional parameters to the channel by passing them through `data` attributes:
   #
   #   <%= turbo_stream_from "room", channel: RoomChannel, data: {room_name: "room #1"} %>
+  #
+  # Raises an +ArgumentError+ if any streamables are blank
+  #
+  #   <%= turbo_stream_from("") %> # => ArgumentError: streamables can't be blank
+  #   <%= turbo_stream_from("channel", nil, "messages") %> # => ArgumentError: streamables can't be blank
```
dhh, line 55: This should imo just collapse to channel/messages. It's fine for something in the streamables array to be nil, just not all of it.

mwallba: @dhh 
Currently streamables like `["channel", nil, "messages"]` or `["channel", nil, "", "messages"]` would collapse to `"channel::messages"` and `"channel:::messages"` when the stream-name is built here:

https://github.com/hotwired/turbo-rails/blob/01ff576bff617af7685e3988a5f8cb2ec553992a/app/channels/turbo/streams/stream_name.rb#L24-L30

I agree that this should just collapse to `"channel:messages"`.
`Turbo::StreamsChannel.signed_stream_name` is a public api though - is it alright to make that change here?

dhh: Hmm, yeah. We could indeed break things like this. Especially with asset caches using the old style. And actually, it doesn't really matter. It's fine to have multiple :::. So I would aim just to raise when nothing but blank is passed.

mwallba: > So I would aim just to raise when nothing but blank is passed.

Sounds good - I've made that change in the most recent commit. Let me know if there is something else missing here and I'll take care of it

PR #665, opened 2024-08-17
Add optional method to TagBuilder for morph

AlexKovynev:
continue of this https://github.com/hotwired/turbo-rails/pull/658

### DIFF
--- a/app/models/turbo/streams/tag_builder.rb
+++ b/app/models/turbo/streams/tag_builder.rb
@@ -77,8 +77,9 @@ class Turbo::Streams::TagBuilder
   #   <%= turbo_stream.replace "clearance_5" do %>
   #     <div id='clearance_5'>Replace the dom target identified by clearance_5</div>
   #   <% end %>
-  def replace(target, content = nil, **rendering, &block)
-    action :replace, target, content, **rendering, &block
+  #   <%= turbo_stream.replace clearance, "<div>Morph the dom target</div>", method: :morph %>
+  def replace(target, content = nil, method: nil, **rendering, &block)
+    action :replace, target, content, method: method, **rendering, &block
   end
 
   # Replace the <tt>targets</tt> in the dom with either the <tt>content</tt> passed in, a rendering result determined
@@ -90,8 +91,9 @@ class Turbo::Streams::TagBuilder
   #   <%= turbo_stream.replace_all ".clearance_item" do %>
   #     <div class='.clearance_item'>Replace the dom target identified by the class clearance_item</div>
   #   <% end %>
-  def replace_all(targets, content = nil, **rendering, &block)
-    action_all :replace, targets, content, **rendering, &block
+  #   <%= turbo_stream.replace_all clearance, "<div>Morph the dom target</div>", method: :morph %>
+  def replace_all(targets, content = nil, method: nil, **rendering, &block)
+    action_all :replace, targets, content, method: method, **rendering, &block
   end
 
   # Insert the <tt>content</tt> passed in, a rendering result determined by the <tt>rendering</tt> keyword arguments,
@@ -155,8 +157,9 @@ class Turbo::Streams::TagBuilder
   #   <%= turbo_stream.update "clearance_5" do %>
   #     Update the content of the dom target identified by clearance_5
   #   <% end %>
-  def update(target, content = nil, **rendering, &block)
-    action :update, target, content, **rendering, &block
+  #   <%= turbo_stream.update clearance, "<div>Morph the dom target</div>", method: :morph %>
+  def update(target, content = nil, method: nil, **rendering, &block)
+    action :update, target, content, method: method, **rendering, &block
   end
 
   # Update the <tt>targets</tt> in the dom with either the <tt>content</tt> passed in or a rendering result determined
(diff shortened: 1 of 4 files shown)


app/models/turbo/streams/tag_builder.rb
```diff
@@ -77,8 +77,8 @@ def remove_all(targets)
   #   <%= turbo_stream.replace "clearance_5" do %>
   #     <div id='clearance_5'>Replace the dom target identified by clearance_5</div>
   #   <% end %>
-  def replace(target, content = nil, **rendering, &block)
-    action :replace, target, content, **rendering, &block
+  def replace(target, content = nil, method = nil, **rendering, &block)
```
mrosso10, line 80: @AlexKovynev 

maybe this should be a keyword argument instead of a positional argument so one can pass rendering options without specifying the method?

btw, thanks for opening this PR :raised_hands:

app/models/turbo/streams/tag_builder.rb
```diff
@@ -229,20 +229,24 @@ def prepend_all(targets, content = nil, **rendering, &block)
   end
 
   # Send an action of the type <tt>name</tt> to <tt>target</tt>. Options described in the concrete methods.
-  def action(name, target, content = nil, allow_inferred_rendering: true, **rendering, &block)
+  def action(name, target, content = nil, method = nil, allow_inferred_rendering: true, **rendering, &block)
     template = render_template(target, content, allow_inferred_rendering: allow_inferred_rendering, **rendering, &block)
 
-    turbo_stream_action_tag name, target: target, template: template
+    turbo_stream_action_tag name, target: target, template: template, **attributes_from_method(method)
```
mrosso10, line 235: I think there's no need for `attributes_from_method`, could simply be replaced by `, method: method`

app/models/turbo/streams/tag_builder.rb
AlexKovynev: keyword with default attributes can't be mixed with ordinal. I tried.

AlexKovynev: DONE but i am not sure that mixing is a good way. Anyway check please

app/models/turbo/streams/tag_builder.rb
AlexKovynev: DONE

mrosso10 (approved):
:+1: :raised_hands:

test/streams/streams_controller_test.rb
```diff
@@ -23,6 +23,7 @@ class Turbo::StreamsControllerTest < ActionDispatch::IntegrationTest
       <turbo-stream action="replace" target="message_1"><template>#{render(message_1)}</template></turbo-stream>
       <turbo-stream action="replace" target="message_1"><template>Something else</template></turbo-stream>
       <turbo-stream action="replace" target="message_5"><template>Something fifth</template></turbo-stream>
+      <turbo-stream method="morph" action="replace" target="message_5"><template>Something fifth</template></turbo-stream>
```
radanskoric, line 26: Maybe also cover the `action="update"` case in exactly the same way?

app/models/turbo/streams/tag_builder.rb
```diff
@@ -77,8 +77,8 @@ def remove_all(targets)
   #   <%= turbo_stream.replace "clearance_5" do %>
   #     <div id='clearance_5'>Replace the dom target identified by clearance_5</div>
   #   <% end %>
-  def replace(target, content = nil, **rendering, &block)
-    action :replace, target, content, **rendering, &block
+  def replace(target, content = nil, method: nil, **rendering, &block)
```
radanskoric, line 80: It's not currently clear from the docs here what the `method` parameter is for. One needs to go to https://turbo.hotwired.dev/reference/streams#replace to understand. Maybe it would be useful to have a example for it in the method docs?

```suggestion
#   <%= turbo_stream.replace clearance, "<div>Morph the dom target</div>", method: :morph %>
  def replace(target, content = nil, method: nil, **rendering, &block)
```

radanskoric (commented):
Thank you so much for opening this, I was **just** looking for it! :)

app/models/turbo/streams/tag_builder.rb
AlexKovynev: DONE

test/streams/streams_controller_test.rb
AlexKovynev: update not covered at all as you see but anyway added with morph :)

radanskoric (approved):
Looks great. 👍

hjhart:
Looks good to me! Eager to use this functionality in my rails project shortly.

OutlawAndy:
Thanks for this PR, @AlexKovynev

Anything in particular holding it up at this point??

AlexKovynev:
> Thanks for this PR, @AlexKovynev
> 
> Anything in particular holding it up at this point??

Nope. Only somebody who can merge. Look like repository support very low now.

AlexKovynev:
@dhh can it me merged somehow please?

jorgemanrubia (approved):
Thanks @AlexKovynev 🙏

jorgemanrubia:
I'll release a new version with this one later today.

Issue #679, opened 2024-09-18
v2.0.8 introduced issue where `broadcast_remove_to` requires `render: false`


JamesChevalier:
v2.0.8 and v2.0.9 contain an issue where `broadcast_remove_to` requires `render: false`

For example:
`after_destroy -> { broadcast_remove_to("#{user_id}-activities") }`
throws "ActionView::MissingTemplate: Missing partial" errors unless it's rewritten to
`after_destroy -> { broadcast_remove_to("#{user_id}-activities", render: false) }`

Introduced in https://github.com/hotwired/turbo-rails/pull/511
Initial commit: https://github.com/hotwired/turbo-rails/commit/5d6e6042d4ff03c0418b7d37e43588f5106e22ec
Rebase commit: https://github.com/hotwired/turbo-rails/commit/76f492a91085e127a9d0ab16e4be7bf1fce33b6a

Code in question:
https://github.com/hotwired/turbo-rails/blob/5d6e6042d4ff03c0418b7d37e43588f5106e22ec/app/models/concerns/turbo/broadcastable.rb#L126-L133

seanpdoyle:
@JamesChevalier thank you for opening this issue. I've opened https://github.com/hotwired/turbo-rails/pull/680 to attempt to resolve the underlying issue.

PR #680, opened 2024-09-18
Ensure `turbo-stream[action="remove"]` does not render a view partial by default

seanpdoyle:
Closes [#679][]

Prior to this change, calls to
`Turbo::Streams::Broadcasts#broadcast_remove_to` would render a view partial that corresponds to the model *regardless* of whether or not the contents of that render would be included in the broadcast `<turbo-stream>` element.

Not only is it wasteful to render HTML and then do nothing with it, it also poses the risk of failing due to errors raised during the rendering process.

This commit restructures the contents of the
`Turbo::Streams::Broadcasts#broadcast_action_to` method to skip extraneous rendering if `render: false` is passed. In addition, it ensures that calls to `broadcast_remove_to` pass `render: false` by default.

[#679]: https://github.com/hotwired/turbo-rails/issues/679

### DIFF
--- a/app/channels/turbo/streams/broadcasts.rb
+++ b/app/channels/turbo/streams/broadcasts.rb
@@ -6,7 +6,7 @@ module Turbo::Streams::Broadcasts
   include Turbo::Streams::ActionHelper
 
   def broadcast_remove_to(*streamables, **opts)
-    broadcast_action_to(*streamables, action: :remove, **opts)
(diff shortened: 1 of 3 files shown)


seanpdoyle:
@jorgemanrubia @brunoprietog are either of you available to review this?

brunoprietog (approved):
Thanks @seanpdoyle!

PR #690, opened 2024-10-03
Add `capture_turbo_stream_broadcast` test helper

seanpdoyle:
Follow-up to [#466][]

Introduce the `#capture_turbo_stream_broadcast` helper to serve as a shortcut invocation for `#capture_turbo_stream_broadcasts`. It returns *a single* value, instead of an Array.

The benefit is that assertions that only need to make an assertion about a single element are not required to deconstruct the `Array` value to access the first element.

[#466]: https://github.com/hotwired/turbo-rails/pull/466

### DIFF
--- a/lib/turbo/broadcastable/test_helper.rb
+++ b/lib/turbo/broadcastable/test_helper.rb
@@ -167,6 +167,48 @@ module Turbo
           document.at("body").element_children
         end
       end
+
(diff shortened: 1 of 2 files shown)


dhh:
Hmm, I'm not sure this is worth it. Maybe we can just add this example:

```ruby
      #     replace = capture_turbo_stream_broadcasts "messages" do
      #       message.broadcast_replace
      #     end.first
```

Or similar.

Issue #507, opened 2023-10-25
Devise FailureApp can't include Turbo::Native::Navigation anymore


arusa:
As learned from @joemasilotti, I was including `Turbo::Native::Navigation` in my Devise FailureApp, so I could use the method `turbo_native_app?` there.

Since v1.5.0 that doesn't work anymore, because Turbo::Native::Navigation uses `helper_method`, which is not available in Devise::FailureApp:

https://github.com/hotwired/turbo-rails/blob/097d8f90cf0c5ed24ac6b1a49cead73d49fa8ab5/app/controllers/turbo/native/navigation.rb#L10

This is not a big issue, as I can easily duplicate the `turbo_native_app?` method.

But I wonder if this is intended, or if classes, that don't know about helpers, should still be able to `include Turbo::Native::Navigation`.

joemasilotti:
Does this work for you?

```ruby
class ApplicationController < ActionController::Base
  helper_method :turbo_native_app?
end
```

seanpdoyle:
@arusa The concern is intended to extend `ActionController::Base` descendants (since it's in defined in `app/controllers/`).

If the maintainers are interested in that kind of broader support, a PR that guarded https://github.com/hotwired/turbo-rails/blob/main/app/controllers/turbo/native/navigation.rb#L10 with a `respond_to?(:helper_method)` block might be enough to implement it.

arusa:
Sounds like an easy solution.

If it's intended to extend `ActionController::Base` there could be more "features" added later that break the apps again, that include it somewhere else.

Maybe the method `turbo_native_app?` should live in a more generic place, where it is accessible from everywhere and `Turbo::Native::Navigation` could also use that.

excid3:
```ruby
class TurboFailureApp < Devise::FailureApp
  # Compatibility for Turbo::Native::Navigation
  class << self
    def helper_method(name)
    end
  end

  include Turbo::Native::Navigation

  # Turbo Native requests that require authentication should return 401s to trigger the login modal
  def http_auth?
    turbo_native_app? || super
  end
end

PR #691, opened 2024-10-04
Migrate Turbo Native helpers to Hotwire Native

joemasilotti:
This PR renames the `turbo_native_app?` helper to `hotwire_native_app?` and changes the user agent check to look for "Hotwire Native".

It is backwards compatible with existing Turbo Native apps by aliasing the `turbo_native_app?` helper and also checking for "Turbo Native" in the user agent.

### DIFF
--- a/app/controllers/turbo/native/navigation.rb
+++ b/app/controllers/turbo/native/navigation.rb
@@ -1,20 +1,23 @@
 # Turbo is built to work with native navigation principles and present those alongside what's required for the web. When you
-# have Turbo Native clients running (see the Turbo iOS and Turbo Android projects for details), you can respond to native
-# requests with three dedicated responses: <tt>recede</tt>, <tt>resume</tt>, <tt>refresh</tt>.
+# have Hotwire Native clients running (see the Hotwire Native iOS and Hotwire Native Android projects for details),
+# you can respond to native requests with three dedicated responses: <tt>recede</tt>, <tt>resume</tt>, <tt>refresh</tt>.
 #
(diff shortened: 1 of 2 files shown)


app/controllers/turbo/native/navigation.rb
```diff
@@ -1,20 +1,23 @@
...
+# you can respond to native requests with three dedicated responses: <tt>recede</tt>, <tt>resume</tt>, <tt>refresh</tt>.
 #
-# turbo-android handles these actions automatically. You are required to implement the handling on your own for turbo-ios.
+# Hotwire Native Android and Hotwire Native iOS handle these actions automatically.
 module Turbo::Native::Navigation
   extend ActiveSupport::Concern
 
   included do
-    helper_method :turbo_native_app?
+    helper_method :hotwire_native_app?, :turbo_native_app?
```
ZilvinasKucinskas, line 10: It breaks the following patch for Devise:

https://github.com/hotwired/turbo-rails/issues/507#issuecomment-1816664359

Quick fix for anyone stumbling across:

```
class TurboFailureApp < Devise::FailureApp
  # Compatibility for Turbo::Native::Navigation
  class << self
    def helper_method(*names)
    end
  end

  include Turbo::Native::Navigation

  # Turbo Native requests that require authentication should return 401s to trigger the login modal
  def http_auth?
    turbo_native_app? || super
  end
end
```

devrizwanali: Thanks @ZilvinasKucinskas :)

PR #696, opened 2024-10-29
Update to rack 3

ramhoj:
This was previously blocked, see https://github.com/rails/rails/pull/46594.

Rails 7.1 now supports rack 3 and Capybara assumes rack 3, making tests fail when run via rake (but not bin/test).


seanpdoyle:
@jorgemanrubia @brunoprietog are either of you able to review this? It restores CI to a passing state for other PRs.

jorgemanrubia (approved):
Thanks @ramhoj!

Issue #669, opened 2024-08-22
Turbo frame request is removing the csrf meta tag


noctivityinc:
When doing a simply turbo frame request via a link click, we are noticing the CSRF meta tag is REMOVED from the page!?  We are having to do this hack to get it back, which I don't love and seems like a bug:

```
document.addEventListener("turbo:before-fetch-request", function (event) {
  window.csrfToken = document.querySelector("[name=csrf-token]").content;
  event.detail.fetchOptions.headers["X-CSRF-Token"] = csrfToken;
});

document.addEventListener("turbo:frame-load", function (event) {
  const csrfMetaTag = document.querySelector('meta[name="csrf-token"]');
  if (!csrfMetaTag) {
    const newCsrfMetaTag = document.createElement("meta");
    newCsrfMetaTag.name = "csrf-token";
    newCsrfMetaTag.content = window.csrfToken;
    document.head.appendChild(newCsrfMetaTag);
  }
});
``

4lllex:
you're going to have to make a reproducible example. a simple turbo frame navigation only updates the frame for me.

sfnelson:
We also ran in to this issue. It seems like a consequence of the change in Turbo 8 to treat frame navigations as full requests (with head). I've spent a lot of time investigating the code paths before for different reasons. See: https://github.com/hotwired/turbo/issues/1047 e.g.

@seanpdoyle I think this is probably something that should be on your radar?

seanpdoyle:
I've opened https://github.com/hotwired/turbo-rails/pull/697 in an attempt to resolve this issue.

jorgemanrubia:
Hey folks I can't reproduce this one. I have tried with both forms and regular links targetting turbo-frames. The `<head>` remains untouched. Could you provide a reproduction case, maybe?

sfnelson:
Our experience of this issue is in an app where we included the turbo tracked elements in the head and use data-turbo-advance on the frame.

https://github.com/hotwired/turbo-rails/pull/697#issuecomment-2702489493

Issue #580, opened 2024-02-14
Turbo Frame Responses missing tracked elements


sfnelson:
When rendering a turbo-frame with `data-turbo-action="advance"` to promote visits, Turbo checks the head of the current document against the head of the incoming response to decide whether to 'render' the synthetic navigation.

In the turbo tests, this check passes because the head of the frame response matches the page response, but in a frame response generated by turbo-rails, the head is empty and so the check fails.

This behaviour changed in Turbo 8/Turbo Rails 2, and as a consequence frame navigation visits are no longer compatible with history restorations.

seanpdoyle:
Is this also related to https://github.com/hotwired/turbo/issues/1047? 

The reproduction in https://github.com/hotwired/turbo-rails/pull/581 demonstrates that https://github.com/hotwired/turbo/pull/1144 might not've resolved the underlying issue.

sfnelson:
Yes, I believe that's related. I've found a workaround for the issue – replacing `layout/turbo_rails/frame.html.erb` and adding the tracked elements to the head element:

```
<head>
  <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
  <%= javascript_importmap_tags %>
  <%= yield head %>
</head>
```

This change might be acceptable as a short term patch to get frame navigations working with `turbo-rails >= 8`, but I think it's likely to cause a lot of headaches:
- Assumes import maps
- Masks the need to add additional lines for any extra turbo-tracked elements
- Assumes that any custom layouts use the same elements
It doesn't seem like a good general solution.

After I made this change, I was running through one of our larger apps updating syntax and I came across a bug that I then spent a good half an hour debugging, only to realise it was this same bug – another layout rendering which doesn't track the application stylesheet – and it was so unclear that this was happening that even though I was literally testing for this change I didn't realise that I was seeing another symptom of this same issue. If the Turbo team is going to stick with the strategy of sending a full `<head>` in turbo frame responses, then I think there needs to be some mechanism for alerting users when the visit is aborted.

sfnelson:
For future me/anyone else debugging this issue, this is what happens when a turbo-frame advance occurs:

1. Frame navigation triggered
2. If this is going to be an _advance_ navigation, then `proposeVisitIfNavigatedWithAction` is called to set up the necessary state tracking and side-effects, this includes taking a whole-page snapshot (saved for later).
[...]

PR #697, opened 2024-10-29
Include CSRF `<meta>` elements in frame layout

seanpdoyle:
Closes [#669][]

If a response to a request with the `Turbo-Frame:` header does not include the `<meta>` elements in the `<html>` document, it's likely that the browser will remove any `<meta>` element present after handling navigating the `<turbo-frame>` that originated the request.

In support of testing this behavior, this commit enables CSRF protection in the test suite.

[#669]: https://github.com/hotwired/turbo-rails/issues/669

### DIFF
--- a/app/views/layouts/turbo_rails/frame.html.erb
+++ b/app/views/layouts/turbo_rails/frame.html.erb
@@ -1,5 +1,6 @@
 <html>
   <head>
+    <%= csrf_meta_tags %>
     <%= yield :head %>
   </head>
   <body>
--- a/test/dummy/app/controllers/application_controller.rb
+++ b/test/dummy/app/controllers/application_controller.rb
@@ -1,2 +1,3 @@
 class ApplicationController < ActionController::Base
+  protect_from_forgery
 end
--- a/test/dummy/config/environments/test.rb
+++ b/test/dummy/config/environments/test.rb
@@ -28,7 +28,7 @@ Rails.application.configure do
   config.action_dispatch.show_exceptions = :none
 
   # Disable request forgery protection in test environment.
-  config.action_controller.allow_forgery_protection = false
+  config.action_controller.allow_forgery_protection = true
 
   # Print deprecation notices to the stderr.
   config.active_support.deprecation = :stderr
--- a/test/frames/frame_request_controller_test.rb
+++ b/test/frames/frame_request_controller_test.rb
@@ -14,6 +14,8 @@ class Turbo::FrameRequestControllerTest < ActionDispatch::IntegrationTest
 
     assert_select "head", count: 1
     assert_select "meta[name=test][content=present]"
+    assert_select "meta[name=csrf-param]"
+    assert_select "meta[name=csrf-token]"
   end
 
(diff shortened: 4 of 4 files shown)


seanpdoyle:
This is a more conservative, less disruptive alternative to https://github.com/hotwired/turbo-rails/pull/534.

While this might resolve any CSRF-related inconsistencies in the `<head>`, it is still an incomplete solution. For example, it does not include [CSP `<meta>` elements](https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/CspHelper.html#method-i-csp_meta_tag), PWA `<meta>` elements, or any other `<head>` elements that [new Rails applications are generated with by default](https://github.com/rails/rails/blob/8c80b02a66fb6f2cf6940027d756bf7d94d6cf53/railties/lib/rails/generators/rails/app/templates/app/views/layouts/application.html.erb.tt#L5-L27).

While #534 might feel too ambitious of a proposal, it circumvents this entire class of issue, and removes any burden of responsibility to decide which `<head>` elements are worth including by default.

ramhoj:
Regarding the failing Rails 7.1 tests, I had the same issue in another branch and submitted https://github.com/hotwired/turbo-rails/pull/696 to address them.

seanpdoyle:
@jorgemanrubia @brunoprietog are either of you experiencing this issue in your applications?

jorgemanrubia:
I don't manage to reproduce this one. In the way turbo works, for turbo-frame requests Turbo won't do any head processing. Why would it remove CSRF tokens from the header? I'll ask for a reproduction case in the original issue.

sfnelson:
@jorgemanrubia in our experience of this issue, we're including turbo-tracked elements in our frame, i.e.

```
<% content_for(:head) do %>
  <%= stylesheet_link_tag :app, "data-turbo-track": "reload" %>
  <%= javascript_importmap_tags %>
<% end %>
```

I can produce a sample application if that is useful, but this was the only change of note for us to reproduce this issue.

sfnelson:
@jorgemanrubia we've been using @seanpdoyle's patch in production since November when we identified the issue. This solution isn't perfect as @seanpdoyle pointed out – it brute force solves CSRF but removes all other meta tags.

We don't want to send all the turbo tracked elements in our frame heads, but if we don't then performing a frame advance visit triggers https://github.com/hotwired/turbo-rails/issues/580 which leaves turbo in an inconsistent state where navigation pop no longer works.

I think there's a wider discussion to be had about how turbo frame advance visits as the status quo is very challenging to navigate successfully.

We'd be happy to contribute time and resources to if this is something that is on the hotwire/turbo team's radar. Our ideal outcome is being able to push a navigation from the server that changes the URL and participates in browser history, but doesn't change the background DOM, similar to a hotwire native modal visit.

jorgemanrubia:
Thanks for the clarifications @sfnelson. I'll merge this one. Maybe a more general solution is what @seanpdoyle proposed in https://github.com/hotwired/turbo-rails/pull/534 (thanks @brunoprietog for pointing that out). But, for now, let's fix this one.

PR #682, opened 2024-09-18
Commit to `ruby@2.6` support

seanpdoyle:
Closes [#681][]

First, remove `ruby@3.0` syntax including `...` arguments and end-less
method definitions.

Next, add `ruby@2.6` and `ruby@2.7` to the CI matrix along with
`rails@7.1`.

[#681]: https://github.com/hotwired/turbo-rails/issues/681

Issue #681, opened 2024-09-18
Ruby 3.0 syntax introduced from v2.0.8 while gem still claims to support Ruby 2.6.0


dentarg:
Happened with https://github.com/hotwired/turbo-rails/commit/5038ed8fcf046f9dbc03b959280a747ee67a08ae#diff-45934cb0c9b4dc0fdd1aa4e780742e07541f16cb60835e1b1330a56d1b0f18f1R9

https://github.com/hotwired/turbo-rails/blob/780ee0da511a3d104683b6852ee00594569109e2/turbo-rails.gemspec#L12

```
$ chruby-exec 2.7 -- ruby -v -e 'def foo() = (@bar)'
ruby 2.7.8p225 (2023-03-30 revision 1f4d455848) [arm64-darwin21]
-e:1: syntax error, unexpected '='
def foo() = (@bar)
-e:1: syntax error, unexpected end-of-input, expecting `end'

$ chruby-exec 3.0.7 -- ruby -v -e 'def foo() = (@bar)'
ruby 3.0.7p220 (2024-04-23 revision 724a071175) [arm64-darwin21]
```

dentarg:
What is the intention, bump `required_ruby_version`? Or was this an accident?

dhh:
We should bump that minimum version, but also, it shouldn't have happened in a tiny version increment. That was a mistake. cc @seanpdoyle

seanpdoyle:
This is a test-only change, and can be converted harm-free.

If 2.6 is the target version, we must add `2.6` and `2.7` to the CI matrix. That can be undone in a follow up release.

seanpdoyle:
While https://github.com/hotwired/turbo-rails/commit/5038ed8fcf046f9dbc03b959280a747ee67a08ae#diff-45934cb0c9b4dc0fdd1aa4e780742e07541f16cb60835e1b1330a56d1b0f18f1R9 is a commit that adds newer 3.0 and 3.1 syntax, reverting those changes alone is not sufficient for executing the test suite for `ruby@2.6` and `rails@6.1` in CI.

There are a variety of gem version incompatibilities and other syntactical incompatibilities. The gem has not been `ruby@2.6` compatible for some time, and there are likely numerous patch and minor versions released since that time.

[...]

PR #698, opened 2024-10-30
Drop Support for `ruby@2.x.x`

seanpdoyle:
Related to https://github.com/hotwired/turbo-rails/issues/681
Follow-up to https://github.com/hotwired/turbo-rails/pull/682

Commit to requiring `ruby@3.1`
    
Remove `ruby@2.6`, `ruby@2.7`, `ruby@3.0`, `rails@6.1`, and `rails@7.0` from the CI matrix, along with any `RUBY_VERSION < "3"` conditionals.
    
The implementation changes include replace older syntaxes with newer variants. For example, utilize `...` in place of `*args, &block`, rely on `**`-ing attribute `Hash` arguments, etc.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,17 +5,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rails: [ "6.1", "7.0", "7.1", "7.2" ]
-        ruby: [ "2.7", "3.0", "3.1", "3.2", "3.3" ]
+        rails: [ "7.1", "7.2", "8.0" ]
+        ruby: [ "3.1", "3.2", "3.3", "3.4" ]
         allow-fail: [ false ]
         include:
-          - { ruby: "2.6",  rails: "6.1" }
           - { ruby: "3.3",  rails: "main", allow-fail: true }
           - { ruby: "3.2",  rails: "main", allow-fail: true }
           - { ruby: "head", rails: "main", allow-fail: true }
         exclude:
-          - { ruby: "2.7", rails: "7.2" }
-          - { ruby: "3.0", rails: "7.2" }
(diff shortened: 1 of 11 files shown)


seanpdoyle:
@jorgemanrubia I've rebased off `main`, and I think this fixes the failing tests. 

Does this changeset achieve the same outcome as https://github.com/hotwired/turbo-rails/tree/fix-tests? If not, maybe you can cherry-pick the changes necessary to pass CI.

jorgemanrubia:
Lovely, thanks @seanpdoyle! I had started to look into fixing the tests, I didn't realize you had this one open 🙏

SleeplessByte:
FYI (for future readers), this being a patch version change broke our build unexpectedly (of course caught by CI, so impact minimal). This release increased the minimum Rails version to 7.1, which is not yet in [Debian (stable)](https://packages.debian.org/source/bookworm/rails).

If you are on Debian 12, pin to [v2.0.12](https://github.com/hotwired/turbo-rails/tree/v2.0.12), or vendor it. In fact you can grab [v2.0.13](https://github.com/hotwired/turbo-rails/tree/v2.0.13) and manually edit the `.gemspec` file, if you're stuck on Debian 12 for now, as Debian ruby is [3.1.2](https://packages.debian.org/bookworm/ruby), and there is no actual requirement (in this repo's code) to run Rails 7.1+, yet. [Trixie has just started the freeze](https://release.debian.org/trixie/freeze_policy.html), which means that Debian 13 will likely release with [Rails 7.2.2.1](https://packages.debian.org/source/trixie/rails).

For those thinking that Rails 6.1 / 7.0 are EOL, the Ruby Debian team is still backporting bugfixes, as is the promise for all things Ruby in Debian.

PR #705, opened 2024-12-23
Update @rails/actioncable dependency to version >=7.0

adrienpoly:
I upgraded to Rails 8.0 our app and noticed that we now have 2 versions of actioncable in our bundle due to the dependency of turbo-rails to actioncable ^7.0

This PR loosen the dependency  to >=7.0


adrienpoly:
I think the CI failures are not directly linked to this PR and are addressed by #706

adrienpoly:
@dhh changes to app/js are the result of runing `yarn build`.

If I don't include them the CI fails

CleanShot 2025-03-04 at 01 23 04@2x

I have rebased main and ran a new `yarn build` to remove conflicts

app/assets/javascripts/turbo.js
```diff
@@ -5533,12 +5533,11 @@ class ConnectionMonitor {
   isRunning() {
     return this.startedAt && !this.stoppedAt;
   }
-  recordPing() {
+  recordMessage() {
```
dhh, line 5536: But this is an actual code change?

app/assets/javascripts/turbo.js
```diff
@@ -5737,6 +5736,7 @@ Connection.prototype.events = {
       return;
     }
     const {identifier: identifier, message: message, reason: reason, reconnect: reconnect, type: type} = JSON.parse(event.data);
+    this.monitor.recordMessage();
```
dhh, line 5739: So is this?

dhh:
Ah, so the code changes are actually pulled from the npm source?

adrienpoly:
> Ah, so the code changes are actually pulled from the npm source?

Yes these changes comes from the build process

PR #710, opened 2025-02-04
Fix unwanted model retention when using class methods from `Turbo::Broadcastable` with locals

v-yarotsky:
Consider the following model:

```ruby
class MyModel < ApplicationRecord
  broadcasts locals: { something_static: true }
end
```

When broadcasting for the first time without this fix, Broadcastable does something like:

```ruby
{ something_static }.reverse_merge!(my_model: #<MyModel: id=1>)
```

On the second broadcast for another model instance, the reverse merge will be a no-op, because the first model is already in the Hash!

Switching to a non-destructive `reverse_merge` resolves the issue.


miharekar:
Could we please get a release with this? 🙏

PR #724, opened 2025-03-13
Turbo::Streams::Broadcasts#render_format support for controller_klass option

ne006:
Rails by default includes all view helpers in a view's context when rendering it, but provides `action_controller.include_all_helpers` option ([docs](https://api.rubyonrails.org/v8.0.0/classes/ActionController/Helpers.html)) to disable such behaviour:
```
config.action_controller.include_all_helpers = false
```

When broadcasting to streams, `Turbo::Streams::Broadcasts#render_format` uses `ApplicationController` to call `render`.
Combined with disabled option above, for example a `broadcast_render_to` call fails when template uses some helpers not included by Rails automatically.

This patch introduces `controller_klass` option for `Turbo::Streams::Broadcasts#render_format` and methods built on top of it to pass a subclass of `ActionController::Base` which includes all the needed helpers and possibly other behaviour needed to successfully render the required template:

```
broadcast_render_to(
      'items',
      template: 'items/update',
      format: ['turbo_stream'],
      locals: { item: self },
      controller: V2::VerySpecificItemController
    )
```

### DIFF
--- a/app/channels/turbo/streams/broadcasts.rb
+++ b/app/channels/turbo/streams/broadcasts.rb
@@ -108,7 +108,7 @@ module Turbo::Streams::Broadcasts
 
   private
     def render_format(format, **rendering)
-      ApplicationController.render(formats: [ format ], **rendering)
+      (rendering.delete(:controller_klass) || ApplicationController).render(formats: [ format ], **rendering)
     end
 
     def render_broadcast_action(rendering)


GuillouuH:
Commentaire de Matthieu sur Basecamp :
- pourquoi le desactiver en local ? on peut pas tester du coup. Et sinon cf [Campfire - Ruby on Rails](https://3.basecamp.com/5224515/buckets/31206811/chats/5745587317@8459347727) le lien de simon juste au dessus
- tu peux rajouter un return early if base_url_extension == "cn" return
- comme ca il reste juste il locale == chine redirect

PR #727, opened 2025-04-01
Fix stream_name calculation inside of thread

brendon:
Calculating the stream_name from within the `ThreadDebouncer` was leading to incorrect stream channel names in my setup where the `global_id` includes the Tennant id. It's a multi-tenant application and the current database name forms part of the `global_id`. Within the `ThreadDebouncer` it would seem to fixate on a particular database connection rather than the connection used by the current request.

Simply calculating the `stream_name` before invoking the `ThreadDebouncer` fixed the issue.

I don't see the need for additional tests unless you wanted to test my failure scenario but I'm not quite sure how to do that.

### DIFF
--- a/app/channels/turbo/streams/broadcasts.rb
+++ b/app/channels/turbo/streams/broadcasts.rb
@@ -69,7 +69,7 @@ module Turbo::Streams::Broadcasts
 
   def broadcast_refresh_later_to(*streamables, request_id: Turbo.current_request_id, **opts)
     stream_name = stream_name_from(streamables)
-    
+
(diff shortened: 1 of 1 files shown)


brendon:
Can I get a review on this one please? So far it's working well for us in production.

jorgemanrubia (approved):
Thanks @brendon

brendon:
Thanks @jorgemanrubia, if it's not too much trouble, would you be able to cut a new release? :)

brendon:
> Thanks @jorgemanrubia, if it's not too much trouble, would you be able to cut a new release? :)

Just following this one up @jorgemanrubia :)

PR #740, opened 2025-05-28
Include hidden elements for turbo-cable-stream-source selector

excid3:
Chrome considers these elements not visible (likely due to their 0px size) so the assertion never matches. This changes the default visible flag to include all elements.

cc @seanpdoyle


lib/turbo/system_test_helper.rb
```diff
@@ -80,6 +80,8 @@ def assert_no_turbo_cable_stream_source(...)
   end
 
   Capybara.add_selector :turbo_cable_stream_source do
+    visible :all
```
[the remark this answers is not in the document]
excid3: Applying display none does get Cuprite to reproduce the issue. 

```erb
<%= turbo_stream_from @message, style: "display: none" %>
```

```
Failure:
AssertionsTest#test_#assert_turbo_cable_stream_source_treats_the_locator_as_:signed_stream_name_filter [test/system/assertions_test.rb:9]:
expected to find visible turbo_cable_stream_source #<Message id: 1, content: nil, created_at: "2025-05-29 13:38:44.716759000 +0000", updated_at: "2025-05-29 13:38:44.716759000 +0000"> 1 time but there were no matches. Also found "", which matched the selector but not all filters.


bin/rails test test/system/assertions_test.rb:4
```

test/system/broadcasts_test.rb
```diff
@@ -123,7 +123,7 @@ class BroadcastsTest < ApplicationSystemTestCase
   private
 
   def reconnect_cable_stream_source(from:, to:)
-    cable_stream_source = find("turbo-cable-stream-source[signed-stream-name=#{signed_stream_name(from)}]")
+    cable_stream_source = find("turbo-cable-stream-source[signed-stream-name=#{signed_stream_name(from)}]", visible: :all)
```
seanpdoyle, line 126: Is this an opportunity to use the selector more directly?


```suggestion
    cable_stream_source = find(:turbo_cable_stream_source, signed_stream_name: from)
```

seanpdoyle (approved):

test/system/broadcasts_test.rb
excid3: Ah yes, great idea.

PR #741, opened 2025-05-29
Add the ability to customize the key used for the instance in the bro…

basabin54:
For certain applications, like when using STI, it can be useful to be able to change the key that is used for the instance in the broadcast locals when trying to reuse partials or displaying tables with a mix of records that are in the same database table but have a different STI subclass. This PR uses a similar approach to broadcast_target_default, preserving the default but allowing it to be easily changed in the model.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ Turbo can coexist with Rails UJS, but you need to take a series of upgrade steps
 
 ## Testing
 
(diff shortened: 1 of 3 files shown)


OutlawAndy:
I believe defining an explicit `self.model_name` method in your STI superclass, is the idiomatic way to resolve the issue you're describing.  The comment on line 519 also says something to that affect.

PR #742, opened 2025-05-29
Add turbo frame assertion test helpers

excid3:
This adds integration test helpers for turbo frames. There are already helpers for turbo streams so this brings Frames into parity with them.

### DIFF
--- a/lib/turbo/test_assertions.rb
+++ b/lib/turbo/test_assertions.rb
@@ -79,5 +79,79 @@ module Turbo
       selector << %([targets="#{targets}"]) if targets
       assert_select selector, count: 0
     end
+
+    # Assert that the rendered fragment of HTML contains a `<turbo-frame>`
+    # element.
+    #
+    # ==== Arguments
+    #
+    # * <tt>ids</tt> [String, Array<String, ActiveRecord::Base>] matches the element's <tt>[id]</tt> attribute
+    #
+    # ==== Options
+    #
+    # * <tt>:loading</tt> [String] matches the element's <tt>[loading]</tt>
+    #   attribute
+    # * <tt>:src</tt> [String] matches the element's <tt>[src]</tt> attribute
+    # * <tt>:target</tt> [String] matches the element's <tt>[target]</tt>
+    #   attribute
+    # * <tt>:count</tt> [Integer] indicates how many turbo frames are expected.
+    #   Defaults to <tt>1</tt>.
+    #
+    #   Given the following HTML fragment:
+    #
+    #     <turbo-frame id="example" target="_top"></turbo-frame>
+    #
+    #   The following assertion would pass:
+    #
+    #     assert_turbo_frame id: "example", target: "_top"
+    #
+    # You can also pass a block make assertions about the contents of the
+    # element. Given the following HTML fragment:
+    #
+    #     <turbo-frame id="example">
+    #       <p>Hello!</p>
+    #     </turbo-frame>
+    #
+    #   The following assertion would pass:
+    #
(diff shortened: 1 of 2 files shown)


lib/turbo/test_assertions.rb
```diff
@@ -79,5 +79,76 @@ def assert_no_turbo_stream(action:, target: nil, targets: nil)
...
+    #       assert_select "p", text: "Hello!"
+    #     end
+    #
+    def assert_turbo_frame(id:, loading: nil, src: nil, target: nil, count: 1, &block)
+      selector = %(turbo-frame[id="#{id}"])
+      selector << %([loading="#{loading}"]) if loading
+      selector << %([src="#{src}"]) if src
+      selector << %([target="#{target}"]) if target
+      assert_select selector, count: count, &block
+    end
```
seanpdoyle, line 124: What do you think about omitting the `count:` keyword option, and instead forwarding all remaining options:

```suggestion
    def assert_turbo_frame(id:, loading: nil, src: nil, target: nil, **options, &block)
      selector = %(turbo-frame[id="#{id}"])
      selector << %([loading="#{loading}"]) if loading
      selector << %([src="#{src}"]) if src
      selector << %([target="#{target}"]) if target
      assert_select selector, **options, &block
    end
```

lib/turbo/test_assertions.rb
```diff
@@ -79,5 +79,76 @@ def assert_no_turbo_stream(action:, target: nil, targets: nil)
...
+    #
+    #     assert_no_turbo_frame id: "example", target: "_top"
+    #
+    def assert_no_turbo_frame(id:, loading: nil, src: nil, target: nil)
+      selector = %(turbo-frame[id="#{id}"])
+      selector << %([loading="#{loading}"]) if loading
+      selector << %([src="#{src}"]) if src
+      selector << %([target="#{target}"]) if target
+      assert_select selector, count: 0
+    end
```
seanpdoyle, line 152: If https://github.com/hotwired/turbo-rails/pull/742/files#r2114782416 feels viable, could this be implemented by forwarding options to `assert_turbo_frame`?

```suggestion
    def assert_no_turbo_frame(**options, &block)
      assert_turbo_frame(**options, count: 0, &block)
    end
```

lib/turbo/test_assertions.rb
```diff
@@ -79,5 +79,76 @@ def assert_no_turbo_stream(action:, target: nil, targets: nil)
...
+    #       <p>Hello!</p>
+    #     </turbo-frame>
+    #
+    #   The following assertion would pass:
+    #
+    #     assert_turbo_frame id: "example" do
+    #       assert_select "p", text: "Hello!"
+    #     end
+    #
+    def assert_turbo_frame(id:, loading: nil, src: nil, target: nil, count: 1, &block)
```
seanpdoyle, line 118: Had you considered changing the `id:` keyword argument to an `id` positional argument so that the assertion's interface mimics the [turbo_frame_tag](https://github.com/hotwired/turbo-rails/blob/main/app/helpers/turbo/frames_helper.rb#L38) helper?

```suggestion
    def assert_turbo_frame(*ids, loading: nil, src: nil, target: nil, count: 1, &block)
    
      id = ids.first.respond_to?(:to_key) ? ActionView::RecordIdentifier.dom_id(*ids) : ids.join('_')
```

lib/turbo/test_assertions.rb
excid3: Since frames required IDs which should be unique, I thought it might make sense to set count to 1. Unspecified I believe it passes if 1+?

PR #748, opened 2025-06-24
Accept symbol, proc, and nil arguments for broadcasts_refreshes

mrrooijen:
Currently, `broadcasts_refreshes` only accepts values in the form of a string or nil, where explicitly passing nil bypasses broadcasting but still triggers the `after_create_commit` callback.

This change allows you to pass a symbol (method name to call) or a proc to evaluate to `broadcasts_refreshes` to produce a stream name dynamically at the instance level `after_create_commit`.

This allows us to, for example, specify a parent (Board) to broadcast refreshes to on `create`, while broadcasting to itself (Column) on `update` and `destroy`:

```rb
class Column < ApplicationRecord
  belongs_to :board
  broadcasts_refreshes :board
end
```

Listen for new `Column` creations for `@board`:

```erb
<%= turbo_stream_from @board %> 
```

Listen for updates and deletion of existing `@column`: 

```erb
<%= turbo_stream_from @column %> 
```

This changes also disables the entire `after_create_commit` hook if `nil` is passed in explicitly, rather than allowing it to run without broadcasting anything.

### DIFF
--- a/app/models/concerns/turbo/broadcastable.rb
+++ b/app/models/concerns/turbo/broadcastable.rb
@@ -210,10 +210,20 @@ module Turbo::Broadcastable
       after_commit -> { broadcast_refresh_later_to(stream.try(:call, self) || send(stream)) }
     end
 
-    # Same as <tt>#broadcasts_refreshes_to</tt>, but the designated stream for page refreshes is automatically set to
-    # the current model, for creates - to the model plural name, which can be overriden by passing <tt>stream</tt>.
+    # Same as <tt>#broadcasts_refreshes_to</tt>, but the designated stream for page refreshes is
+    # automatically set to the current model. For creates, the stream defaults to the model's plural
+    # name, which can be overridden by passing <tt>stream</tt>.
+    # If <tt>stream</tt> is a Symbol, a method with that name will be called to determine the stream.
+    # If <tt>stream</tt> is a Proc, it will be called with the model instance to determine the stream.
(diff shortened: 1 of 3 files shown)


app/models/concerns/turbo/broadcastable.rb
```diff
@@ -210,10 +210,20 @@ def broadcasts_refreshes_to(stream)
...
+    # If <tt>stream</tt> is explicitly set to nil, the after_create_commit callback will be skipped entirely.
     def broadcasts_refreshes(stream = model_name.plural)
-      after_create_commit  -> { broadcast_refresh_later_to(stream) }
+      after_create_commit do
+        case stream
+        when String then broadcast_refresh_later_to(stream)
+        when Symbol then broadcast_refresh_later_to(send(stream))
+        else broadcast_refresh_later_to(stream.try(:call, self))
+        end
+      end if stream
```
seanpdoyle: > where explicitly passing `nil` bypasses broadcasting but still triggers the `after_create_commit` callback.

What is the intended behavior of supporting explicitly `nil` arguments? To me, invoking the `after_create_commit` seems to be an accidental side-effect.

Rather than suffixing the `if stream` conditional, would an `ArgumentError` make sense?

```suggestion
      raise ArgumentError, "stream is required…" if stream.nil?
      
      after_create_commit do
        case stream
        when String then broadcast_refresh_later_to(stream)
        when Symbol then broadcast_refresh_later_to(send(stream))
        else broadcast_refresh_later_to(stream.try(:call, self))
        end
      end
```

PR #754, opened 2025-08-25
feat: accept meta tag string arguments as well

frenkel:
Before this PR you would get this error:
`Information for cause: ArgumentError (Invalid refresh option 'morph')`
for this code:
`<%= turbo_refresh_method_tag "morph" %>`

This error message was a real head scratcher until I read the code and realized I had used the wrong type. I think strings are fine to accept as well, this PR fixes it.


seanpdoyle:
@frenkel could you add some tests to cover these values?

seanpdoyle (approved):

seanpdoyle:
Thank you!

PR #756, opened 2025-09-12
Allow `turbo-cable-stream-source` to be compatible with `data-turbo-permanent`

bensheldon:
When the `turbo-cable-stream-source` element is disconnected from the DOM, ~~wait a tick~~ wait until the Turbo Render operation finishes to allow any potential DOM re-connect to occur. This prevents unnecessary channel unsubscribe/subscribe cycles when disconnected and then immediately reconnected with `data-turbo-permanent` and Turbo Drive.

This is possible because all of the Turbo Permanent / [bardo.js](https://github.com/hotwired/turbo/blob/main/src/core/bardo.js) operations are synchronous.

Related issues:
- https://github.com/hotwired/turbo/issues/868
- https://github.com/hotwired/turbo/issues/471
- https://github.com/hotwired/turbo/issues/230
- [An Out of Body Experience with Turbo](https://mileswoodroffe.com/articles/out-of-body-experience-with-turbo)... which makes me wonder if this entire class of problem could be sidestepped by [_moving_](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements#lifecycle_callbacks_and_state-preserving_moves) permanent elements into a holding area in the dom, then _moving_ them into the new page rather than _removing_ and _adding_ them ([a little related in here](https://github.com/hotwired/turbo/pull/305#issuecomment-879282187) and [here](https://github.com/hotwired/turbo/issues/499))


bensheldon:
@seanpdoyle I pushed up a fix for the failing test. I didn't notice that `attributeChangedCallback` was calling `connectedCallback + disconnectedCallback`; I extracted the subscribe/unsubscribe behavior separate from the newly introduced delay.

bensheldon:
Thinking aloud, I'm wondering how we can make the lifecycle more resilient. Waiting for the nextTick works currently, but is dependent on all of the turbo permanent dom operations being synchronous, which maybe isn't a good future-forward expectation. 

I think we could potentially track using turbo render events. As the order of events  seems to be:

1. `turbo:before-render`
2. start turbo permanent operation
3. element is disconnected from dom
4. element is re-connected to dom
5. finish turbo permanent operation 
6. `turbo:render`

I'll take a swing at that, and now that I'm a bit more familiar with the test suite, I'll see if I can lock those in too. 

This turned out to be more fun and interesting than expected 😉

brunoprietog:
> > this entire class of problem could be sidestepped by [moving](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements#lifecycle_callbacks_and_state-preserving_moves) permanent elements into a holding area in the dom, then moving them into the new page rather than removing and adding them ([hotwired/turbo#305 (comment)](https://github.com/hotwired/turbo/pull/305#issuecomment-879282187))
> 
> This is a really interesting idea! The introduction of [Element.moveBefore](https://developer.mozilla.org/en-US/docs/Web/API/Element/moveBefore) and the [connectedMoveCallback](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements#custom_element_lifecycle_callbacks) could be extremely valuable (for both morphing and `[data-turbo-permanent]`).

That's very interesting and would address the root cause of the problem. Currently, permanent elements in Turbo are not truly permanent, as they are detached and attached to the dom anyway. We can't use `moveBefore`, but we have Idiomorph for that, which uses it when available.

PR #760, opened 2025-11-15
TurboStreamEncoder: redefine `accept_header` method

viralpraxis:
Since the `action_dispatch_integration_test` hook is invoked for each integration test, the whole `TurboStreamEncoder` is (re)defined many times.

Ideally, I think, `@encoders[:turbo_stream]` should be set once, but since `RequestEncoder` does not expose straightforward API to check if an encoder is registered, we'd have to use something like

```ruby
ActionDispatch::RequestEncoder::IdentityEncoder === ActionDispatch::RequestEncoder.encoder(:turbo_stream)
```

which is not handy.

NB: requiring AS is not necessary -- we can use the self-alias trick instead


viralpraxis:
any thoughts on this? using method redefinition is also safer since it ensures that redefined method actually exists (that's not the case right now)

PR #761, opened 2025-11-17
Use an immediate debouncer for tests

djmb:
The background debouncer can cause flaky tests as the debounced action could take place during the following test leading to unexpected broadcasts.

Instead let's use an immediate debouncer in tests so that we synchronously broadcast and there's no chance for interference between tests.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -23,13 +23,8 @@ class ActiveSupport::TestCase
   end
 
   def with_production_debouncer(&block)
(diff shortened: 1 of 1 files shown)


dhh:
We can drop Ruby 3.1 as its EOL.

seanpdoyle:
The replacement of 3.1 with 3.2 is still failing CI: https://github.com/hotwired/turbo-rails/actions/runs/19575460052/job/56059415505.

PR #786, opened 2026-06-04
Use a method-based frame layout to avoid _layout_from_proc redefinition warning

### LINKED ISSUE #783: Ruby 4.0 warning: method redefined _layout_from_proc in FrameRequest
## Description

Running on Ruby 4.0.1 with Rails 8.1.3, turbo-rails 2.0.23 produces warnings about `_layout_from_proc` being redefined. This occurs each time a controller inherits from a base that already has a layout set.

## Warnings

```
actionview-8.1.3/lib/action_view/layouts.rb:311: warning: method redefined; discarding old _layout_from_proc
turbo-rails-2.0.23/app/controllers/turbo/frames/frame_request.rb:24: warning: previous definition of _layout_from_proc was here
```

The warning appears multiple times (once per controller that includes the module).

## Cause

`Turbo::Frames::FrameRequest` uses a proc-based layout in its `included` block:

```ruby
included do
  layout -> { "turbo_rails/frame" if turbo_frame_request? }
end
```

Rails' `layout` with a proc triggers `_write_layout_method`, which defines `_layout_from_proc`. When the including controller (or its parent) already has a layout defined, the method gets redefined, triggering the warning.

## Environment

- Ruby 4.0.1
- Rails 8.1.3
- turbo-rails 2.0.23

tachyurgy:
## What

`Turbo::Frames::FrameRequest` installs its frame layout with a Proc:

```ruby
layout -> { "turbo_rails/frame" if turbo_frame_request? }
```

A Proc layout makes Rails' `_write_layout_method` define a private `_layout_from_proc` on every controller that includes the module (it re-runs on each subclass via the `inherited` hook). That `define_method` isn't silenced, so when a controller in the chain also sets its own layout the method gets redefined and Ruby warns:

```
actionview/lib/action_view/layouts.rb:311: warning: method redefined; discarding old _layout_from_proc
turbo-rails/app/controllers/turbo/frames/frame_request.rb:24: warning: previous definition of _layout_from_proc was here
```

It's a verbose-mode warning today, but verbose is **on by default in Ruby 4.0**, so it now surfaces for every app on the upgrade path. Reported in #783.

## Fix

Use a Symbol (method) layout instead of a Proc:

```ruby
layout :turbo_frame_request_layout

# ...
def turbo_frame_request_layout
  "turbo_rails/frame" if turbo_frame_request?
end
```

This resolves identically. In `_write_layout_method`, the Symbol and Proc branches both return the layout name when the value is truthy and fall through to the same default-layout lookup when it's `nil` — but the Symbol branch never defines `_layout_from_proc`, so the warning is gone with no behavior change.

## Test

Adds a regression test asserting the frame layout no longer installs `_layout_from_proc` on including controllers. It fails on the old Proc-based layout and passes with this change. The existing frame-layout behavior tests (minimal layout on frame requests, `head` content, override, unique etag) continue to pass.

Fixes #783

### DIFF
--- a/app/controllers/turbo/frames/frame_request.rb
+++ b/app/controllers/turbo/frames/frame_request.rb
@@ -21,13 +21,24 @@ module Turbo::Frames::FrameRequest
   extend ActiveSupport::Concern
 
   included do
-    layout -> { "turbo_rails/frame" if turbo_frame_request? }
+    layout :turbo_frame_request_layout
     etag { :frame if turbo_frame_request? }
 
     helper_method :turbo_frame_request?, :turbo_frame_request_id
   end
 
   private
+    # Use a method-based (Symbol) layout rather than a Proc. A Proc-based layout
+    # makes Rails define a private +_layout_from_proc+ method on every including
+    # controller, which is then redefined whenever a controller in the chain sets
+    # its own layout — emitting a "method redefined; discarding old
+    # _layout_from_proc" warning under Ruby's verbose mode (on by default in Ruby
+    # 4.0). A Symbol layout resolves the same way (frame layout on frame requests,
+    # default layout otherwise) without defining that method. See issue #783.
+    def turbo_frame_request_layout
+      "turbo_rails/frame" if turbo_frame_request?
+    end
+
     def turbo_frame_request?
       turbo_frame_request_id.present?
     end
--- a/test/frames/frame_request_controller_test.rb
+++ b/test/frames/frame_request_controller_test.rb
@@ -57,6 +57,22 @@ class Turbo::FrameRequestControllerTest < ActionDispatch::IntegrationTest
     end
 end
 
+class Turbo::FrameRequestLayoutTest < ActiveSupport::TestCase
+  # Regression test for #783. The frame layout must be installed as a method
+  # (Symbol) rather than a Proc. A Proc-based layout makes Rails define a private
+  # `_layout_from_proc` on every controller that includes the module, which is
+  # then redefined whenever a controller in the chain sets its own layout —
+  # emitting a "method redefined; discarding old _layout_from_proc" warning under
+  # verbose mode (on by default in Ruby 4.0).
+  test "frame layout does not define _layout_from_proc on including controllers" do
+    controller = Class.new(ActionController::Base)
+
+    assert_not controller.private_instance_methods.include?(:_layout_from_proc),
+      "Turbo's frame layout should be a Symbol-based layout so that Rails does " \
+      "not define (and later redefine, with a warning) _layout_from_proc. See #783."
+  end
+end
+
 class Turbo::FrameRequestViewTest < ActionView::TestCase
   test "supports rendering context without request object" do
     @rendered = ApplicationController.render(template: "trays/show")


tachyurgy:
Gentle bump on this one. CI is still green across all five matrix entries (Ruby 3.2/3.4, Rails 7.2 through main) and GitHub still reports it as cleanly mergeable, so it has not gone stale.

Recapping for anyone picking it up cold: the warning in #783 comes from the proc-based `layout` in `frame_request.rb`, which Ruby re-defines as `_layout_from_proc` and then warns about on redefinition. This swaps it for a method-based layout. The Symbol and Proc branches resolve identically, so behavior is unchanged, and there is a regression test that fails on the old code and passes on the new.

Happy to rebase, split it up, or take it in a different direction if the maintainers would rather solve the warning another way. Just let me know.

test/frames/frame_request_controller_test.rb
```diff
@@ -57,6 +57,22 @@ def with_prepended_view_path(path, &block)
...
+  # `_layout_from_proc` on every controller that includes the module, which is
+  # then redefined whenever a controller in the chain sets its own layout —
+  # emitting a "method redefined; discarding old _layout_from_proc" warning under
+  # verbose mode (on by default in Ruby 4.0).
+  test "frame layout does not define _layout_from_proc on including controllers" do
+    controller = Class.new(ActionController::Base)
+
+    assert_not controller.private_instance_methods.include?(:_layout_from_proc),
+      "Turbo's frame layout should be a Symbol-based layout so that Rails does " \
+      "not define (and later redefine, with a warning) _layout_from_proc. See #783."
```
seanpdoyle, line 72: Could this be replaced with an [assert_not_includes](https://edgeapi.rubyonrails.org/classes/ActiveSupport/TestCase.html#method-i-assert_not_includes) call?


```suggestion
    assert_not_includes controller.private_instance_methods, :_layout_from_proc,
      "Turbo's frame layout should be a Symbol-based layout so that Rails does " \
      "not define (and later redefine, with a warning) _layout_from_proc. See #783."
```

app/controllers/turbo/frames/frame_request.rb
```diff
@@ -21,13 +21,24 @@ module Turbo::Frames::FrameRequest
...
   end
 
   private
+    # Use a method-based (Symbol) layout rather than a Proc. A Proc-based layout
+    # makes Rails define a private +_layout_from_proc+ method on every including
+    # controller, which is then redefined whenever a controller in the chain sets
+    # its own layout — emitting a "method redefined; discarding old
+    # _layout_from_proc" warning under Ruby's verbose mode (on by default in Ruby
+    # 4.0). A Symbol layout resolves the same way (frame layout on frame requests,
+    # default layout otherwise) without defining that method. See issue #783.
```
seanpdoyle, line 37: This contribution includes:

1. Git commit message outlining the rationale
2. Pull request description and thread outlining the rationale
3. Code comment on the private method (I'm not sure that private methods generate RubyDoc)
4. Code comment before a `test "…"` declaration in a test file outlining the rationale
5. Assertion message outlining the rationale that will be presented in case of making a change that accidentally induces a test failure

Is there a way to choose one or two of these occurrences to include, then omit the rest?

My preference is to retain the test assertion message (to guard against future regressions) as well as the git commit. What do you think?

seanpdoyle: Otherwise, this change looks good to me. Thank you for this contribution!

